PR #631, opened 2023-05-14
Playwright e2e tests

oznasi1:


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
+on:
+  push:
+    branches: [main, master]
```
zavelevsky, line 4: we don't have master branch - why keep it?

e2e/auth.setup.ts
```diff
@@ -0,0 +1,41 @@
+import { test as setup } from '@playwright/test';
+
```
zavelevsky, line 2: test files should be named accordingly e.g. .spec.ts or test.ts

src/libs/modals/Modal.tsx
```diff
@@ -49,6 +49,7 @@ export const Modal: FC<Props> = ({
       className={'px-content items-center justify-center'}
     >
       <m.div
+        id="modal"
```
zavelevsky, line 52: WAT

oznasi1:
Check

Issue #345, opened 2023-02-27
Create strategy - warning when creating empty strategy


ashachaf:
as a user, i would like to be warned when i create a strategy with no budget (on both tokens).

## Flow
1. create strategy
2. input rates
3. DO NOT input any budget 
4. click "create strategy"
expect:
add a popup to confirm this is not a mistake (which will end up costing more gas to the user when he needs to fund the strategy).
**on the popup**
- confirm -> proceed with the create strategy flow and trigger wallet signature
- reject -> close the popup and show the create strategy page (with all previously filled data)

## Mockup
https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=4126%3A66043&t=3HkW0KIJC2mE3XXZ-4
https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=4126%3A66217&t=3HkW0KIJC2mE3XXZ-4

PR #656, opened 2023-05-25
#345 - Add create strategy no budget warning

pingustar:
#345


ashachaf:
CTA color is incorrect
image

pingustar:
> CTA color is incorrect image

changed the cancel button to black because modal background color was changed which caused this problem with the secondary variant button.

ashachaf (approved):

Issue #649, opened 2023-05-22
Multiple language support


Costa769:
1. Prepare the infrastructure for multiple language support
1.1. Define rules for long text handling such as decreasing the font size and/or breaking the text into two lines in case its above a certain length threshold.
2. Provide a single file with all the Carbon app texts to the product team.
2.2. Translate the texts
3. Implement the translations and language selection in UI
4. Website display language will be determined by the language of the user's browser.
4.1 In case we do not support the browser language of the user, the website will be displayed in English
4.2 In case user changes the language of the website, his choice will be stored in cookies so the next time the website is loaded with the language that was selected by the user.
5. The URL of the website won't change if the user changes the language. For example the default app URL https://app.carbondefi.xyz/ will remain the same if user changes the language from EN to ES.

Designs:
https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=6547%3A109964&t=QWn7tpEkNU5BMkgI-1
https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=6560%3A111314&t=QWn7tpEkNU5BMkgI-1

Reference:
https://github.com/bancorprotocol/webapp-v2/tree/3b367a51d8e5c3c8472e539fe1d5ccfe2d3708a4/src/i18n
https://github.com/bancorprotocol/webapp-v3/tree/fe0f77c4893b85a9ee1ae6e153f37fe8c5a6b787/src/i18n

Costa769:
> @Costa769 @ashachaf Do we have the design for languages manu?

Added the designs

PR #658, opened 2023-05-28
#649: multiple language support

oznasi1:
Added infra to support multiple languages using i18next library - https://www.i18next.com/.
That POC contains the option to change language between English (default) and Spanish using the burger menu.

Also, we are supporting detecting user language, load it, store it in local storage and if the user change the language manually, local storage is updating and in the next entrance to the app, we will get the language from there.

For now, I'm using i18next Backend which allows us to import on demand only the language that we use and not all of them at ones. that's why I wrapped the app with Suspense.

Also, we have a cache based on local storage with versioning and expiration time of 7 days. The flow is: 
App is loaded, check local storage available with same version - getting translation from there. if local storage does not exist so we get the translation and update local storage. And of course if version is not updated we are getting the translation from translation file and update local storage.

Also, for each translation key, Iv'e added default value to protects us in case of missing key or some other unknown error.


public/locales/en-US/translation.json
```diff
@@ -0,0 +1,20 @@
+{
+  "common": {
```
pingustar, line 2: I wouldnt use common attributes for the translations. 

I think you will run into unwanted side effects when changing text that does not apply to all "common" occurences.

public/locales/en-US/translation.json
```diff
@@ -0,0 +1,20 @@
...
+    "cancelButton": "Cancel"
+  },
+  "modal": {
+    "delete": {
```
pingustar, line 6: modal.delete is not very descriptive. What are we deleting? Currently we have only strategies but lets prepare for other possible features in the future where we would have another delete modal.

src/components/core/menu/mainMenu/MainMenuRightBurger/useBurgerMenuItems.tsx
```diff
@@ -8,6 +8,8 @@ import { ReactComponent as IconTelegram } from 'assets/logos/telegram.svg';
 import { ReactComponent as IconV } from 'assets/icons/v.svg';
 import { useFiatCurrency } from 'hooks/useFiatCurrency';
 import { MenuItemActions } from './useMenuContext';
+import { useTranslation } from 'libs/translations';
+import { SUPPORTED_LANGUAGES } from 'languages/i18n';
```
pingustar, line 12: I would expect this file to be in the `libs/translations` folder

src/index.tsx
```diff
@@ -1,17 +1,19 @@
...
+import { App } from 'App';
+import { StoreProvider } from 'store';
 import { Web3ReactWrapper } from 'libs/web3';
 import { Router } from 'libs/routing';
 import { LazyMotion } from 'libs/motion';
 import { QueryProvider } from 'libs/queries';
-import { StoreProvider } from 'store';
+import { CarbonLogoLoading } from 'components/common/CarbonLogoLoading';
+import './languages/i18n';
+import 'global-shim';
```
pingustar, line 12: why are you reordering these imports? I am pretty sure there was a reason that global shim and sentry is at the top.

src/index.tsx
```diff
@@ -24,7 +26,17 @@ root.render(
         <Web3ReactWrapper>
           <LazyMotion>
             <Router>
-              <App />
+              <Suspense
```
pingustar, line 29: why do we introduce suspense for i18n support? Is it required/related?

src/libs/modals/modals/ModalConfirmStrategy/ModalConfirmStrategy.tsx
```diff
@@ -63,7 +65,7 @@ export const ModalConfirmStrategy: ModalFC<ModalConfirmStrategyData> = ({
         <IconTitleText
           variant={data?.variant}
           icon={data?.icon}
-          title={data?.title || ''}
+          title={data?.subtitle || ''}
```
pingustar, line 68: is this intentional? 

title vs. subtitle

src/libs/translations/index.ts
```diff
@@ -0,0 +1,8 @@
+import HttpBackend from 'i18next-http-backend';
```
pingustar, line 1: cant you export directly?

`export HttpBackend from "i18next-http-backend"`

oznasi1: Yes, wanted to import as default and name export so I used that:
export { default as HttpBackend } from 'i18next-http-backend';

Applied to all imports in the file.

src/libs/modals/modals/ModalConfirmStrategy/ModalConfirmStrategy.tsx
oznasi1: it was, but renamed it now to be more clear.

src/components/core/menu/mainMenu/MainMenuRightBurger/useBurgerMenuItems.tsx
```diff
@@ -174,10 +180,33 @@ export const useBurgerMenuItems = () => {
     []
   );
 
-  menuMap.set('main', { items: mainItems });
+  const languagesItems = useMemo(
+    (): MenuItemType[] =>
+      SUPPORTED_LANGUAGES.map(({ code, name }) => {
```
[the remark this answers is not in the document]
oznasi1: @pingustar  Do you think the right way is to split all menu items to their own hook?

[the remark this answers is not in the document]
oznasi1: @pingustar After discussion with Jan, that is good for now but be aware for further additional to that file any consider to split this file out.

.eslintrc.json
```diff
@@ -0,0 +1,14 @@
+{
```
zavelevsky, line 1: please move all eslint config here from the package.json

src/components/common/forwardArrow.tsx
```diff
@@ -0,0 +1,29 @@
...
+export const ForwardArrow: FC<ForwardArrowProps> = ({
+  arrowType = 'cut',
+  className = '',
+}) => {
+  const { i18n } = useTranslation();
+  const newClassName = `h-12 w-7 ${
+    i18n.dir() === 'rtl' ? 'rotate-180' : ''
+  } ${className}`;
+
+  return (
```
zavelevsky, line 20: why not 
```js
return arrowType === 'cut' ? <IconArrowCut className={newClassName} /> : <IconArrowFull className={newClassName} />
```

src/components/common/walletConnect/utils.tsx
```diff
@@ -0,0 +1,38 @@
+import { ReactComponent as IconWallet1 } from 'assets/icons/wallet-1.svg';
+import { ReactComponent as IconWallet2 } from 'assets/icons/wallet-2.svg';
+import { ReactComponent as IconWallet3 } from 'assets/icons/wallet-3.svg';
+import { ReactComponent as IconWallet4 } from 'assets/icons/wallet-4.svg';
+import { ReactComponent as IconWallet5 } from 'assets/icons/wallet-5.svg';
+import { ReactComponent as IconWallet6 } from 'assets/icons/wallet-6.svg';
+import { i18n } from 'libs/translations';
+
+export const getProductDescriptionItems = () => {
```
zavelevsky, line 9: utils isn't a good name for this file

oznasi1: amm, is data.tsx sounds better ?

oznasi1: @zavelevsky can u take a look at this again pls?

src/components/common/TokenInputField/TokenInputField.tsx
```diff
@@ -33,11 +34,12 @@ export const TokenInputField: FC<Props> = ({
   isError,
   className,
   onKeystroke,
-  placeholder = 'Enter Amount',
+  placeholder = i18n.t('common.placeholders.placeholder2'),
```
zavelevsky, line 37: so why use the hook at all?

oznasi1: As we talked, we are now using t function from hook only and pass it down to functions if needed.

src/components/common/walletConnect/utils.tsx
oznasi1: Changed to items.tsx

src/components/core/menu/mainMenu/MainMenuRightBurger/useBurgerMenuItems.tsx
```diff
@@ -25,87 +27,96 @@ const iconStyles = 'h-32 w-32 md:h-20 md:w-20';
...
-      },
-      {
-        subMenu: 'resources',
-        content: 'Resources',
-      },
-      {
-        content: (
-          <Link className="flex" to={externalLinks.faq}>
-            FAQ
+  const mainItems: MenuItemType[] = [
```
zavelevsky, line 36: you removed useMemo in a couple of places - why?

src/libs/modals/modals/ModalGenericInfo.tsx
```diff
@@ -52,15 +53,15 @@ export const ModalGenericInfo: ModalFC<ModalGenericInfoData> = ({
         }}
         className={'my-16'}
       >
-        {buttonLabel}
+        {buttonLabel || t('modals.genericInfo.actionButtons.actionButton1')}
```
zavelevsky, line 56: note that you don't allow me to pass an empty string - becuase this check will always override it. You do it in other places too. Not a big issue - but still.

Costa769 (approved):

ashachaf:
to support language on analytics:
change the way we pass page referral back to analytics.
current:
https://github.com/search?q=repo%3Abancorprotocol%2Fcarbon-app%20page_referrer_spa&type=code

updated:
```
page:
 {
    referrer_spa: "/about", //where "/about" is dynamic based on the viewed page
    language: "en"
  }
```

Issue #665, opened 2023-05-31
Create strategy - add popular tags and default values


Costa769:
as a user, i would like to create a strategy that is typical and mostly created (almost as the system default)

the goal is to help users pick the strategy type that suites them knowing that range type strategies require a bit more learning and are more suited for advanced users.

### Designs:
https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=4704%3A89031&t=0vCMYlJ7YCuFthEB-1

https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=6194%3A103801&t=wWW5u1XIpzzdjCCi-1

### adding popular indication
1. Recurring strategies - Add "Popular" tag to the "2 Limits" strategy
2. Disposable strategies- Add "Popular" tag to "Buy Limit" and "Sell Limit" strategies
  2.1 Rearrange strategy options under disposable strategies to display them in the following order: Buy Limit, Sell Limit, Buy Range, Sell Range

### pre-selecting defaults
on the strategy type, we would like to pre-select a default type.
- recurring type:
-- since recurring is the default type, pre-select the "2 Limits" option as a default value (which means the CTA will be enabled as a result)
- disposable type:
-- the "Buy Limit" option should be selected by default.

PR #679, opened 2023-06-12
#665: Add default values for strategy type & recommended label

oznasi1:


src/components/strategies/create/CreateStrategyTypeMenu.tsx
```diff
@@ -30,7 +30,7 @@ const BlockIconTextDesc = ({
       </div>
       <div className={'flex-shrink space-y-6'}>
         <div className={'text-14 font-weight-500'}>{title}</div>
-        <div className={'h-32 text-12 text-white/60'}>{description}</div>
```
pingustar, line 33: why remove the hight class? Now the UI is broken - the icon and buttons jump around when switching between "Recurring" and "Disposable"

src/components/strategies/create/CreateStrategyTypeMenu.tsx
```diff
@@ -49,6 +49,16 @@ export const CreateStrategyTypeMenu: FC<UseStrategyCreateReturn> = ({
     selectedTabItems,
   } = useCreateStrategyTypeMenu(base?.address!, quote?.address!, strategyType);
 
+  useEffect(() => {
```
pingustar, line 52: not sure we need an effect for this. Why cant you set this as default value in the state?

pingustar (changes requested):

src/components/strategies/create/CreateStrategyTypeMenu.tsx
oznasi1: We have some case that we need the useEffect for:
1. user open the page, recurring as default and  first item selected -> click again on Recurring -> the first item won't be selected.
2. user open the page, recurring as default and  first item selected -> click on 2 Ranges -> click on Recurring again the first item won't be selected.

Update me if its not clear so we can have a short discord :)

ashachaf (approved):

pingustar (approved):
LGTM

Issue #666, opened 2023-05-31
Range pricing popup


Costa769:
## Implement a range orders/pricing popup

When a user selects the "range" price options , we want to inform and educate about range prices before proceeding.
Display a popup when user selects a range order type in the strategy creation page or range pricing in the set prices/edit prices page. Once user has clicked on the popup's CTA button, the data will be stored in cookies and the popup won't be displayed again.

### Design:

https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=6250-106422&t=qRVZhoUbzYrDSSKk-4

### Logic
IF cookie exists, do not show the popup
else,
**Create strategy page**
Display the popup when:
- User clicks on "2 Ranges"  or "Custom" under recurring strategies tab
- User clicks on "Buy Range" or "Sell Range" options under disposable strategies tab

**Set Prices/Edit Prices page**
Display the popup when:
- User switches from limit to range pricing
- User comes to the Set prices page via deeplink with range pricing option pre-selected.

***Popup logic***
#### Consent
default setting is no-consent or empty
- IF user clicks the confirm CTA in the popup:
-- store user confirmation
- IF user clicks to close the popup (outside the popup or x)
-- do not save cookie value
#### Behavior
- IF user click CTA in popup -> close popup and select the range option 
- IF user click outside popup or x -> close popup and select the default limit option of that screen (2 limits, limit buy)
- IF user came via deep link and clicked outside popup or x -> close popup and select the default limit option of that screen. In case of the set prices screen, the selected limit price option will have no input in the price field
- IF user duplicated an existing range strategy and click outside popup or x->close popup and selected limit price option with no input in the price field

PR #680, opened 2023-06-12
#666 - Add strategy range warning message

pingustar:
#666


src/components/strategies/create/useCreateStrategyTypeMenu.tsx
```diff
@@ -51,99 +55,117 @@ export const useCreateStrategyTypeMenu = (
...
+  );
 
   const handleClick = (
     to: string,
     search: StrategyCreateLocationGenerics['Search'],
     replace?: boolean
   ) => {
-    navigate({ to, search, replace });
+    if (
+      search.strategySettings !== 'limit' &&
```
oznasi1, line 160: I noticed that in CreateStrategyOrders.tsx you have the opposite condition: 
strategySettings === 'range' || strategySettings === 'custom'
I think it would be better to pick one style and use it in both places . same result but a little bit easier to read.

oznasi1 (changes requested):
@pingustar one case that is not functioning like the task description - maybe something is not updated. In case of duplicate range strategy according to task it should open the modal and if the user close the modal, setting the orders to limit.
Currently this is not the behavior, if we duplicate range strategy the modal not opened.

oznasi1 (approved):
LGTM

ashachaf (approved):

Issue #672, opened 2023-06-04
Create strategy/Edit prices - add market price indication and default values


Costa769:
### Market price difference indication
We want to display the difference between the input price in both buy and sell fields and the current market price

### Logic:
Display the indications on both set prices and edit prices pages 

**Buy price:**
--Display the indication only if the buy price is above market price

**Sell price:**
--Display the indication only if the sell price is below market price

**General:**
--The difference between the input price and the market price will be shown in percentage
--Display percentage with 2 decimals, in case the difference is less than 0.01%, display <0.01%
--Thresholds are not implemented, meaning the indication will be displayed in the same color, regardless of the value

### Add default input values

On create strategy ->set prices page we would like to set default prices in the buy/sell input fields

**Buy price**
Set default buy price 0.5% below the market price

**Sell price**
Set default sell price 0.5% above the market price

In case we don't know the market price?

PR #683, opened 2023-06-18
#672: Create MarketPriceIndication component

oznasi1:


src/components/strategies/create/BuySellBlock/InputLimit.tsx
```diff
@@ -1,10 +1,12 @@
-import { useFiatCurrency } from 'hooks/useFiatCurrency';
-import { Token } from 'libs/tokens';
 import { ChangeEvent, FC, FocusEvent } from 'react';
 import { carbonEvents } from 'services/events';
+import { useFiatCurrency } from 'hooks/useFiatCurrency';
```
pingustar, line 3: why keep reordering imports?

src/components/strategies/create/BuySellBlock/InputLimit.tsx
```diff
@@ -1,10 +1,12 @@
...
+import { Token } from 'libs/tokens';
 import { sanitizeNumberInput } from 'utils/helpers';
 import { decimalNumberValidationRegex } from 'utils/inputsValidations';
 import { ReactComponent as IconWarning } from 'assets/icons/warning.svg';
+import { MarketPriceIndication } from 'components/strategies/marketPriceIndication';
+import { MarketPricePercentage } from 'hooks/useMarketIndication';
 
 export const InputLimit: FC<{
```
pingustar, line 11: lets extract a Props type

src/components/strategies/create/BuySellBlock/InputRange.tsx
```diff
@@ -1,9 +1,11 @@
 import { ChangeEvent, FC, FocusEvent } from 'react';
-import { Tooltip } from 'components/common/tooltip/Tooltip';
+import BigNumber from 'bignumber.js';
+import { carbonEvents } from 'services/events';
+import { Token } from 'libs/tokens';
```
pingustar, line 4: why reorder the imports? For me this is just noise in the review

src/components/strategies/create/BuySellBlock/InputRange.tsx
```diff
@@ -62,7 +69,7 @@ export const InputRange: FC<{
 
   return (
     <div>
-      <div className="flex space-x-6">
+      <div className="grid grid-cols-2 gap-6">
```
pingustar, line 72: why change to grid over flex?

src/components/strategies/create/BuySellBlock/InputLimit.tsx
```diff
@@ -1,10 +1,12 @@
...
-import { Token } from 'libs/tokens';
 import { ChangeEvent, FC, FocusEvent } from 'react';
 import { carbonEvents } from 'services/events';
+import { useFiatCurrency } from 'hooks/useFiatCurrency';
+import { Token } from 'libs/tokens';
 import { sanitizeNumberInput } from 'utils/helpers';
 import { decimalNumberValidationRegex } from 'utils/inputsValidations';
 import { ReactComponent as IconWarning } from 'assets/icons/warning.svg';
+import { MarketPriceIndication } from 'components/strategies/marketPriceIndication';
+import { MarketPricePercentage } from 'hooks/useMarketIndication';
```
pingustar, line 9: I dont think the global hooks folder is the correct place for this file. Do we plan on using this outside of create strategy scope?

src/components/strategies/create/BuySellBlock/index.tsx
```diff
@@ -134,7 +140,15 @@ export const BuySellBlock: FC<Props> = ({
       )}
 
       <LimitRangeSection
-        {...{ base, quote, order, buy, title, inputTitle, isOrdersOverlap }}
+        {...{
```
pingustar, line 143: what changed here? Did the formatter rules change? This looks like more noise in the review

src/components/strategies/marketPriceIndication/index.tsx
```diff
@@ -0,0 +1,53 @@
+import { FC } from 'react';
+import BigNumber from 'bignumber.js';
+import { Tooltip } from 'components/common/tooltip/Tooltip';
+
+type Props = {
+  marketPricePercentage: BigNumber;
+  isRange?: boolean;
```
pingustar, line 7: why is `isRange` optional? In what case do we pass undefined?

src/components/strategies/marketPriceIndication/index.tsx
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export const MarketPriceIndication: FC<Props> = ({
```
pingustar, line 10: why suddenly use index.ts for the component?

Usually our approach is that we have an index file with exports incase we have a folder with multiples files. The index file will indicate which files are exposed outside.

src/hooks/useMarketIndication.ts
```diff
@@ -0,0 +1,83 @@
...
+    if (order.isRange) {
+      const isInputNotZero = buy
+        ? new BigNumber(order.max).gt(0)
+        : new BigNumber(order.min).gt(0);
+
+      return (
+        isInputNotZero &&
+        new BigNumber(getFiatValue(buy ? order.max : order.min))[
+          buy ? 'gt' : 'lt'
+        ](tokenMarketPrice)
```
pingustar, line 42: this is really hard to read and understand whats going on here, lets clean this up a bit.

src/components/strategies/create/useCreateStrategy.ts
```diff
@@ -46,7 +47,20 @@ export const useCreateStrategy = () => {
   const order1 = useOrder(templateStrategy?.order1);
   const order0 = useOrder(templateStrategy?.order0);
   const [isProcessing, setIsProcessing] = useState<boolean>(false);
+  const { marketPricePercentage: buyMarketPricePercentage } =
+    useMarketIndication({
```
pingustar, line 51: it seems to me we are using this hook in to many places. Can you give some context as to why we need this in different places?

src/components/strategies/create/BuySellBlock/InputRange.tsx
```diff
@@ -16,6 +18,10 @@ export const InputRange: FC<{
   buy?: boolean;
   error?: string;
   setRangeError: (error: string) => void;
+  marketPricePercentages: {
+    min: BigNumber;
+    max: BigNumber;
```
pingustar, line 23: why not use the imported type `MarketPriceIndication`?

src/hooks/useMarketIndication.ts
```diff
@@ -0,0 +1,83 @@
...
+import { useFiatCurrency } from './useFiatCurrency';
+import { OrderCreate } from 'components/strategies/create/useOrder';
+
+export type MarketPricePercentage = {
+  min: BigNumber;
+  max: BigNumber;
+  price: BigNumber;
+};
+
+export type UseMarketIndicationProps = {
```
pingustar, line 14: why export this? Do we need this outside this file?

src/hooks/useMarketIndication.ts
```diff
@@ -0,0 +1,83 @@
...
+  base: Token | undefined;
+  quote: Token | undefined;
+  order: OrderCreate;
+  buy?: boolean;
+};
+
+export const useMarketIndication = ({
```
pingustar, line 21: lets discuss this hook, I dont think this is a hook that should be places in the global hooks folder. And I have a feeling we are over complicating things because this data is used only in create strategy, right?

I see the hook beeing used these two files within the same scope:

`src/components/strategies/create/BuySellBlock/LimitRangeSection.tsx`
`src/components/strategies/create/useCreateStrategy.ts`

pingustar (changes requested):

src/App.test.tsx
```diff
@@ -1,4 +1,3 @@
-import React from 'react';
```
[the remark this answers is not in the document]
oznasi1: it's redundant. we don't need to import react in tests files.

src/components/strategies/create/BuySellBlock/InputLimit.tsx
oznasi1: will stop doing it manually. Maybe will add in the future eslint rule for it

src/components/strategies/create/BuySellBlock/InputRange.tsx
oznasi1: You right, missed that. thanks

src/components/strategies/create/useCreateStrategy.ts
oznasi1: of course, we need the market indication for both create strategy & edit strategy.
LimitRangeSection component is common for both of them so I'm using over there.
Also I needed market price indication for gtm events only for create strategy event, that is why I also added the custom hook also to useCreateStrategy. let me know if that make sense to you.

src/hooks/useMarketIndication.ts
oznasi1: pls see the comment above - we need this hook for both edit & create strategy

src/components/strategies/create/BuySellBlock/InputLimit.tsx
oznasi1: you right, moving it to strategy scope. we are using it for both edit & create strategy

src/components/strategies/create/BuySellBlock/InputRange.tsx
pingustar: nice, good job

src/components/strategies/create/BuySellBlock/index.tsx
oznasi1: When added new props it automatically change the format, but when removing it is was still that way.
anyway, fixed that now.

src/components/strategies/marketPriceIndication/index.tsx
oznasi1: Fixed that. Create index.ts file to export all of the needed component,hooks,types. Created also MarketPriceIndication.tsx to contain the component  & now the hook is inside that folder.

src/components/strategies/marketPriceIndication/index.tsx
oznasi1: MarketPriceIndication component is in use from 2 places only: 
InputLimit.tsx & inputRange.tsx so we know if we are in a isRange case or not. 
So, if you don't pass isRange prop, default would be false.

pingustar (approved):
LGTM

Costa769 (approved):

Issue #715, opened 2023-07-18
Tokens appear without logo and with the "suspicious" warning


zavelevsky:
We noticed in production that some token logos are missing.
Only workaround was to clear local storage.

It appears that the function `fetchTokenData`, which marks tokens as suspicious, is used not only for the import token flow but also for other flows and I suspect this leads to the issue.

If this is indeed the case, note that the fix must also include handling existing data in local storage

turbo: 0xA35923162C49cF95e6BF26623385eb431ad920D3
$big: 0xc8De43Bfe33FF496Fa14c270D9CB29Bda196B9B5
<img width="1506" alt="image" src="https://github.com/bancorprotocol/carbon-app/assets/40430285/dd41f44b-4a1e-4791-9cd4-4c2f49349cbf">
<img width="402" alt="image" src="https://github.com/bancorprotocol/carbon-app/assets/40430285/3b9e938a-0740-4c12-b254-ff6156d1e078">

[localstorage.txt](https://github.com/bancorprotocol/carbon-app/files/12078147/localstorage.txt)

PR #735, opened 2023-08-08
#715 - fix token list and import token flow

pingustar:
#715


pingustar:
in order to ship the bug fix now, feedback provided by @zavelevsky will be handled in #769

PR #756, opened 2023-08-29
#747 - Revert translation changes

tiagofilipenunes:
Relevant changes and questions, to facilitate reviewing:
- src/components/common/selectToken/index.tsx
  - [x] Removed useMemo, should it be added?
- src/components/core/menu/mainMenu/MainMenuRightBurger/useBurgerMenuItems.tsx
  - [x] useMemo removed by Oz for resourcesItems, re-add?
- package.json
  - [x] Should the remaining devDependancies be removed? devDependancies were added as part of the translation PR, apart from @types/ua-parser-js
- src/index.tsx
  - [x] Remove Suspense?
- src/components/strategies/edit/EditStrategyHeader.tsx
  - [x] Left change that added unnecessary OR {(type && titleByType[type]) || ''}. Remove?

Changes to check are okay:
- src/components/common/walletConnect/items.tsx
  - Changed function to get items to an item list instead
- src/components/core/menu/menuItems.ts
  - Instead of useMenuItems
- src/components/strategies/overview/strategyBlock/utils.ts
  - Change function to get tooltip text to a tooltip text list instead
  - 'No Budget · Inactive' to 'No Budget - Inactive', to match 'Paused - Inactive'
- src/libs/notifications/data.ts
  - Change function to get notifications to a notification map instead
- src/libs/modals/modals/WalletModal/ModalWallet.tsx
  - Unnecessary check that selectedConnection is null, since isError would be false in that case
- src/libs/notifications/data.ts
  - Typos "(...) have failed." to "(...) has failed."
  - Typo "withdraw request" -> "withdrawal request"


tiagofilipenunes:
PR with the changes being reverted: #658

.eslintrc.json
tiagofilipenunes, line 1: Was getting carriage return error before making this change.

src/components/core/menu/mainMenu/MainMenuRightBurger/useBurgerMenuItems.tsx
```diff
@@ -133,7 +118,7 @@ export const useBurgerMenuItems = () => {
               <span>{currency}</span>
               <span className="flex items-center">
                 <IconV
-                  className={`invisible h-12 w-12 me-8 ${
+                  className={`invisible h-12 w-12 ${
```
pingustar, line 121: just making sure this margin was removed on purpose? I assume it was added because of RTL support?

src/components/trade/settings/TradeSettingsRow.tsx
```diff
@@ -108,7 +106,7 @@ export const TradeSettingsRow: FC<{
           </Button>
         ))}
         <input
-          placeholder={t('common.placeholders.placeholder5')}
+          placeholder={'custom'}
```
pingustar, line 109: should this be lowercase?

vitest.config.ts
```diff
@@ -1,9 +1,8 @@
 import { defineConfig } from 'vitest/config';
-import tsconfigPaths from 'vite-tsconfig-paths';
 
 export default defineConfig({
   test: {
     environment: 'jsdom',
   },
-  plugins: [tsconfigPaths()],
+  plugins: [],
```
pingustar, line 7: hmm lets see what we needed this for, might be worth keeping this plugin

pingustar (approved):
LGTM - great work!

src/components/core/menu/mainMenu/MainMenuRightBurger/useBurgerMenuItems.tsx
tiagofilipenunes: Removed on purpose as it was added by the #658 PR for RTL support

src/components/trade/settings/TradeSettingsRow.tsx
tiagofilipenunes: It's the same as the current value we have for it in the app. 

image

I think it looks best as 'Custom', do you agree?

vitest.config.ts
tiagofilipenunes: Was added as part of the #649 PR, happy to keep it.

I don't see any paths defined in tsconfig.json for which we'd need this vite path mapping plugin though

pingustar (approved):
LGTM

PR #763, opened 2023-09-01
Windows issues

GrandSchtroumpf:
Fixes two issues related to windows during installation : 
- `"endOfLine": "auto"`: fixes the `Delete 'CR'` warn on every end of line
- use escaped double quote in package.json because single quotes are interpreted as a folder name on windows


zavelevsky (approved):
works fine on Mac - so LGTM

pingustar (approved):
LGTM

PR #765, opened 2023-09-01
improve a11y on navigation

GrandSchtroumpf:
Small PR to get to know the code. 
Improve a11y for navigation : 
- use `nav` with `aria-label` to get implicit `navigation` role ([Documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/nav))
- use `aria-current="page"` on anchors for the main page & `aria-current="location"` for tabbed pages. ([Documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current))
- removed a div that had no impact on the UI


zavelevsky (approved):
LGTM

PR #770, opened 2023-09-04
A11y Search

GrandSchtroumpf:
- [x] Use `<input type="search">` instead of type text will have implicit `role="searchbox"` ([Documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/searchbox_role))
- [x] Remove default cancel button from webkit
- [x] Wrap search into a `<form>` with `<button type="reset">` & `<button type="submit">`
- [x] Add labels or aria-label on the `input` as `placeholder` is not recognized by adaptive technologies
- [x] ~Use upcoming `search` tag for future compatibility ([Documentation](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/search)) with `role="search"` on it for backward compatibility ([Documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/search_role))~

-> React throw an error with tag `search` (even if valid HTML) so I set `role="search"` on the `form`


GrandSchtroumpf:
There is some issue remaining with the keyboard navigation on the combobox: onblur closes the widget.
This might take more time to handle, so I'll do it in another PR.

src/components/explorer/ExplorerSearchInput.tsx
```diff
@@ -4,40 +4,25 @@ import { cn, wait } from 'utils/helpers';
...
-      return props.setSearch('');
-    }
-    if (e.key !== 'Enter') {
-      return;
-    }
-    if (!props.isError && props.search.length > 0) {
-      if (props.type === 'token-pair' && props.filteredPairs.length > 0) {
-        const slug =
-          `${props.filteredPairs[0].baseToken.symbol}-${props.filteredPairs[0].quoteToken.symbol}`.toLowerCase();
-        props.setShowSuggestions(false);
```
pingustar, line 28: removing line 24 to 29 introduces the following regressions:

- on enter the first suggestion is not selected
- on enter the suggestions don't disappear

pingustar (changes requested):
regressions  introduced, see comment

GrandSchtroumpf:
- [x] Fix regression
- [x] Support keyboard navigation `ArrowDown`, `ArrowTop`, `Home`, `End`
- [x] Support keyboard (`Enter`, `Espace`) and mouse selection (click & reset)
- [x] Add aria for `comboxbox` & `listbox`
- [x] Increase hover color to differentiate from keyboard selection

I split the input into two components because of they started to diverged with new aria attribute and specific keydown operations.
I created a explorer.module.css file to share styles between inputs & popups

src/components/explorer/ExplorerSearchSuggestions.tsx
```diff
@@ -1,59 +1,191 @@
...
+  KeyboardEvent,
+  memo,
+  SetStateAction,
+  useId,
+  useRef,
+  useState,
+} from 'react';
 import { cn } from 'utils/helpers';
+import { ReactComponent as IconClose } from 'assets/icons/times.svg';
+import styles from './explorer.module.css';
```
pingustar, line 15: lets not use css modules. This is defeating the biggest advantage of tailwind which is not to have to come up with a bunch of different classnames and it makes it harder to understand the component because I need to switch files to see what styles are applied. It also increases css bundle size.

Use a reusable component instead.

src/components/explorer/ExplorerSearchSuggestions.tsx
```diff
@@ -1,59 +1,191 @@
...
 
-interface Props
-  extends Pick<ExplorerSearchProps, 'filteredPairs' | 'setSearch'> {
-  setShowSuggestions: Dispatch<SetStateAction<boolean>>;
+interface Props extends Omit<ExplorerSearchProps, 'type'> {}
+
+function isOption(el?: Element | null): el is HTMLElement {
```
pingustar, line 19: lets stay consistent with the current code style of using arrow functions.

pingustar (changes requested):

src/components/explorer/ExplorerSearchSuggestions.tsx
GrandSchtroumpf: As discussed on Telegram, aria-selected is not supported with our current version of Tailwind. We'll keep this part as CSS for now remove it once we update to latest Tailwind version (#774)
The other classes have been moved to a ContainerComponent.

GrandSchtroumpf:
Work on requested changes:
- [x] Move `onChange` on each input
- [x] Move Suggestion components & utils functions into separate files under `suggestion` folder
- [x] Rename `popupClasses` to `suggestionClasses`
- [x] Put back 0 Results + use h4 for title on SuggestionEmpty

Issue #671, opened 2023-06-04
Fiat currency display rules


Costa769:
Change the way we display fiat values on the website for all the supported currencies

### Current:

If currency = USD, CAD or AUD
Display  "$" symbol before the number and the currency code after the number
Else
Display the currency code after the number

**Examples:**

USD: $100 USD
AUD: $100 AUD
CAD: $100 CAD
EUR: 100 EUR
JPY: 100 JPY

### Updated:
**IF** one of the fiat currencies is selected (USD, EUR ,GBP, JPY, AUD, CAD, CHF and CNH)
User the mapping table below to display a currency symbol & currency code before the amount:

USD-$ 
EUR-€
GBP-£
JPY-¥
AUD-A$
CAD-CA$
CHF-CHF
CNH-CNH

**Examples:**
USD: $100
EUR: €100
GBP: £100
JPY:  ¥100
AUD: A$100
CAD: CA$100
CHF: CHF 100
CNH: CNH 100

**ELSE** (if crypto currency is selected)
Display the currency code after the number

**Examples:**
ETH:  100 ETH

GrandSchtroumpf:
We can use `Int.NumberFormat` for that ([Documentation](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat))

PR #777, opened 2023-09-07
671 currency display

tiagofilipenunes:
#671

- [x] Consider using numbro(1000.234).formatCurrency() or [Intl.NumberFormat](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat)
- [x] Collect locale standards
	- https://en.wikipedia.org/wiki/IETF_language_tag
	- https://en.wikipedia.org/wiki/ISO_3166-1
	- https://en.wikipedia.org/wiki/ISO_4217
- [x] Check spec format is correct against ISO standards
	- [x] All codes exist
	- [x] All currency formatting is consistent with the standards
	- [x] JPY - no digits after the decimal separator
	- [x] Intl.NumberFormat respects ISO4217 and the spec provided
- [x] Change prettifyNumber
- [x] Update tests and add new ones for EUR, JPY and ETH

# Comments for review
- Vitest is using jsdom which doesn't support Intl functionality. This functionality is provided by the local node.js installation, so when testing locally, make sure to use node version >=18
- Intl.supportedValuesOf() is only available from Typescript 5.1.3, and current version is 4.4.2. It is [widely supported](https://caniuse.com/mdn-javascript_builtins_intl_supportedvaluesof)
- roundingMode in the Intl.NumberFormatOptions is not specified in the intl.d.ts types and will only be supported on future versions of typescript. The option is [widely supported though](https://caniuse.com/mdn-javascript_builtins_intl_numberformat_numberformat_options_roundingmode_parameter).
- There is a loss of precision from converting num:BigNumber to num:number. It should be fine for currency numbers, but not for others.

# Equivalence of terms

| Numbro   format                         | IntlFormat                                                                        |
| --------------------------------------- | --------------------------------------------------------------------------------- |
| average = true                          | notation = compact                                                                |
| rounded                                 | RoundingMode = 'halfExpand'                                                       |
| floored                                 | RoundingMode = floor                                                              |
| mantissa x                              | Rounding type = fractionDigits, MaximumFractionDigits = MinimumFractionDigits = X |
| optional mantissa                       | minimumFractionDigits = 0                                                         |
| trim mantissa                           | minimumFractionDigits = 0                                                         |
| low precision (doesn't exist in Numbro) | Not-applicable                                                                    |
| space separated                         | 1k -> 1 k, not natively supported                                                 |
| thousand separated                      | useGrouping = true (default)                                                      |

## Abbreviate
- Numbro Options
	- average true => 1234 = 1k
	- if round is true, round the number, otherwise floor it
	- space separated 1k -> 1 k
	- low precision option doesn't exist in numbro but is specified
	- mantissa 1
	- optional mantissa
- Intl.NumberFormat:
	- notation: 'compact'
	- round is default behaviour roundingMode: "halfExpand", floor is roundingMode: "floor"
	- can't do space separated
	- optional mantissa& mantissa 1 = minimumFractionDigits: 1

## Low Precision
- Numbro Options
	- if round is true, round the number, otherwise floor it
	- mantissa 0
	- thousand separated -> 1234 = 1,234
	- optional mantissa
	- trim mantissa
- Intl.NumberFormat:
	- round is default behaviour "halfExpand", floor is roundingMode "floor"
	- thousand separated is useGrouping: true
	- optional mantissa&trim mantissa&mantissa 0 = maximumFractionDigits: 0

## Fallback currency formatting option (default behaviour)
- Numbro Options
	- if round is true, round the number, otherwise floor it
	- mantissa 2
	- thousand separated -> 1234 = 1,234
	- optional mantissa
	- trim mantissa
- Intl.NumberFormat:
	- round is default behaviour "halfExpand", floor is roundingMode "floor"
	- useGrouping: true
	- optional mantissa&trim mantissa & mantissa 2 = default behaviour


pingustar (approved):
Code LGTM.

Issue #771, opened 2023-09-04
Explorer - token pair issue for tokens with dash in symbol


ashachaf:
when trying to search for tokens that include dash in their symbol, the explorer fails to return valid results

example:
https://app.carbondefi.xyz/explorer/token-pair/eth-eth2x-fli
token pair should be:
token0: eth
token1: eth2x-fli

current: as you can see, we show 0 results for this pair.
expected: show the relevant strategies for this pair.
#### NOTE: we might need to remove support for dash in the search. currently, we identify dash as token separator

for reference, trade shows there is liquidity for this pair
https://app.carbondefi.xyz/trade?base=0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE&quote=0xAa6E8127831c9DE45ae56bB1b0d4D4Da6e5665BD

pingustar:
note: same problem appears when searching for a trade pair with a dash in the symbol

GrandSchtroumpf:
As a side effect of this issue, when we select `eth2x-fli/eth` it's transformed into `eth2x/fli-eth` in the search box.

PR #778, opened 2023-09-07
#771 - Fix token symbol with dash issue

pingustar:
#771


pingustar:
closed in favour of #780

PR #781, opened 2023-09-07
501 reinstall husky

tiagofilipenunes:
#501


tiagofilipenunes:
When a commit is created:
- The staged files (and only the staged files) are modified with eslint --fix for errors only, using lint-staged
- yarn test and yarn build run consecutively
- If there are any errors when running the hook, the commit is cancelled


To quickly test the pre-commit hook behaviour locally
1. yarn install to install husky hooks in your local machine
2. Modify any file and add an import that will remain unused
3. Add a random line of code in the file that would not trigger a linting error
4. Stage the file, check the diff
5. Create a commit, yarn test and yarn build are also ran. 
6. Compare the diff in the commit changes vs the diff in step 3 - the unused import was removed

pingustar (approved):
LGTM

e2e/utils/tenderly.ts
```diff
@@ -1,5 +1,6 @@
 const TENDERLY_ACCESS_KEY = process.env['TENDERLY_ACCESS_KEY'];
-if (!TENDERLY_ACCESS_KEY) throw new Error('No TENDERLY_ACCESS_KEY provided in .env');
+if (!TENDERLY_ACCESS_KEY)
+  throw new Error('No TENDERLY_ACCESS_KEY provided in .env');
```
pingustar, line 3: nit: lets wrap this in brackets to improve readability

package.json
```diff
@@ -92,14 +94,15 @@
...
+    "postinstall": "yarn compile-abis && yarn husky install",
     "analyze": "source-map-explorer 'build/static/js/*.js'",
     "start": "vite",
     "build": "tsc && vite build",
     "serve": "vite preview",
     "test": "vitest run",
     "e2e": "playwright test",
-    "lint": "eslint --ext .tsx,.ts ."
+    "lint": "eslint --ext .tsx,.ts .",
+    "pre-commit": "yarn lint-staged && yarn test"
```
pingustar, line 105: I think we can also skip testing and only run the linter @zavelevsky what do you think?

@tiagofilipenunes regarding `yarn lint-staged` I only see a `yarn lint` command for eslint.

Where is lint-staged? Shouldnt it be lint?

And is `eslint --ext .tsx,.ts .` sufficient or do we need to run prettier format in addition to this?

tiagofilipenunes: I also think we can skip testing, since it happens in our CI

lint-staged will run linting on staged files, modify them, and stage those changes, all before the commit. 

it uses the .lintstagedrc.json, and runs eslint, then runs prettier

e2e/utils/tenderly.ts
tiagofilipenunes: Done

package.json
zavelevsky: > @zavelevsky removed testing from pre-commit hook. Looks good now?

yes

Issue #779, opened 2023-09-07
Explorer Search regression


pingustar:
when deep linking to a non existing pair, the search suggestions are showing 0 results when clearing the incorrect input.

steps to reproduce:

1. go to link: https://app.carbondefi.xyz/explorer/token-pair/eth-usdcccc
2. select all in search input and remove

expected: suggestions should show all results

actual: suggestions show 0 results

https://github.com/bancorprotocol/carbon-app/assets/7262066/70b9aa1c-a796-45aa-82fd-94fecbd95a7a

PR #782, opened 2023-09-07
Fix suggestion regression on first delete

GrandSchtroumpf:
#779 
- [x] update `search` on input instead of change in the combobox


pingustar (approved):
LGTM

PR #788, opened 2023-09-08
E2e/playwright

GrandSchtroumpf:
- [x] Setup playwright
- [x] Global setup create a new fork, update the debug page and save local storage
- [x] First test goes to strategy page, wait for content to be visible and take a screenshot
- [x] Global teardown delete current fork
- [x] Github action push new commit
- [x] Add two env flags CI & DRAFT to take screenshots only if CI && !DRAFT

Current behavior:
- Create a new fork on global setup
- e2e tests runs on all changes (Draft or not)
- Screenshot are done only on CI && not draft
- Delete fork on global teardown


e2e/setup.ts
```diff
@@ -0,0 +1,49 @@
+import { chromium, firefox, webkit, type FullConfig } from '@playwright/test';
+import { CreateForkBody, createFork, forkRpcUrl } from './utils/tenderly';
+
+const forkConfig: CreateForkBody = {
```
zavelevsky, line 4: As we said, we probably need to fork from a specific block. Yes, using a dedicated wallet reduces the chance of unexpected results - but for every other needs (e.g. testing the explorer) you want to get the same data.

e2e/setup.ts
```diff
@@ -0,0 +1,49 @@
...
+async function globalSetup(config: FullConfig) {
+  console.log('Setting up Tenderly Fork');
+  console.time('Fork is setup');
+
+  // Create a Fork
+  const fork = await createFork(forkConfig);
+  process.env['TENDERLY_FORK_ID'] = fork.id;
+  const rpcUrl = forkRpcUrl(fork.id);
+  // Binance address
+  const imposter = '0x47ac0Fb4F2D84898e4D9E7b4DaB3C24507a6D503';
```
zavelevsky, line 21: should be env var and kept out of git log

e2e/setup.ts
```diff
@@ -0,0 +1,49 @@
...
+  const setupProjects = config.projects.map(async (project) => {
+    if (!(project.name in browsers)) return;
+    const { baseURL, storageState } = project.use;
+    if (!baseURL) return;
+
+    const browser = await browsers[project.name as BrowserName].launch();
+    const page = await browser.newPage();
+    await page.goto(`${baseURL}/debug`);
+    // SET RPC-URL
+    await page.getByLabel('RPC URL').fill(rpcUrl);
```
zavelevsky, line 33: Page Object pattern is preferable IMO

e2e/setup.ts
GrandSchtroumpf: The `forkConfig` has an params `blockNumber` to specify. I was waiting to discuss it with you during Sync. I should have left a comment about that sorry.

e2e/setup.ts
GrandSchtroumpf: We discussed this in sync and couldn't find any reason to put it in env. As tests will rely on this address, so if each dev change it in his env, it might impact the tests.
Why do you think it should be in env ?

e2e/teardown.ts
```diff
@@ -0,0 +1,13 @@
+import { deleteFork } from './utils/tenderly';
+
+async function globalTeardown() {
```
pingustar, line 3: no action required, but lets try to keep arrow syntax for consistency in the future

pingustar (approved):
LGTM

e2e/setup.ts
zavelevsky: As discussed - no other constraint - just pick what's current at the moment

e2e/setup.ts
zavelevsky: as discussed - this is a very minor concern - resolving comment

Issue #795, opened 2023-09-10
Fix: Intl.supportedValuesOf not available on some browser engine versions


tiagofilipenunes:
On some browser versions, supportedValuesOf might not be available.

https://caniuse.com/mdn-javascript_builtins_intl_supportedvaluesof

PR #796, opened 2023-09-10
Check if supportedValuesOf is available

tiagofilipenunes:
#795


tiagofilipenunes:
Can polyfill with @formatjs/intl-enumerator. However, Intl.supportedValuesOf is supported by 89% of browsers, so checking whether it exists is also a good way to understand if Intl.NumberFormat will format the currency number adequately

pingustar (approved):
LGTM

PR #798, opened 2023-09-11
Default sort to Roi descending

GrandSchtroumpf:
fix #791
- [x] default compare fn is roi descending
- [x] default localstorage is roi descending
- [x] fallback localstorage is roi descending

PR #800, opened 2023-09-11
Replace strategy sort & filter enum

GrandSchtroumpf:
String literal are more readable than enum in :
- Codebase
- URL searchParams
- GTM events

As I had to send sort/filter to GTM I thought it would be better to start with this migration to string literals.

- [x] Create mapping between sort/filter and there labels (Ok because we removed i18n)
- [x] Create type from this mapping
- [x] Use mapping for sortFn instead of switch
- [x] Create localstorage migration from enum to string literal
- [x] Use "roiDesc" as default (as #798 is not yet merged)


zavelevsky (approved):
LGTM

PR #801, opened 2023-09-11
Add explorer events

### LINKED ISSUE #792: Tracking - updated event tracking
https://docs.google.com/spreadsheets/d/1NPZ-NtmFwJhW0v1bgmdSz8X7MbGmsJOa70Hi_po-cKo/edit#gid=1439640141

GrandSchtroumpf:
fix #792
blocked by #800
- [x] Add explorer event types & handler
- [x] Implement events

Concerning the `explore_search_results_filter_sort` requires the value for search, filter and sort on filter/sort changes.
These changes happen on the store, but search is not there. I see two solutions
1. On `setFilter`/`setSort`  check if we are in `/explorer` with `window.location` and query input.
```typescript
const setSort = (sort: StrategySort) => {
  _setSort(sort);
  lsService.setItem('strategyOverviewSort', sort);
  if (window.location.pathname.startsWith('/explorer')) {
    const search = document.querySelector<HTMLInputElement>('input[type="search"]').value;
    explorerEvents.exploreSearchResultsFilterSort({ search, sort, filter });
  }
};
```
2. In the `ExplorerSearch` had another wrapper around `setFilter`/`setSort`:
```typescript
const { strategies } = useStore();
const { sort, filter } = strategies;
const setSort = (sort: StrategySort) => {
  explorerEvents.exploreSearchResultsFilterSort({ search, sort, filter });
  strategies.setSort(sort);
};
```

I went with option 2, but I can also use option 1 if you prefer.


pingustar (approved):
Code LGTM. Check with product / analytics if everything is good on their end.

src/components/strategies/overview/strategyBlock/StrategyBlockManage.tsx
```diff
@@ -43,13 +45,14 @@ export const StrategyBlockManage: FC<Props> = ({
   setManage,
   isExplorer,
 }) => {
+  const { strategies } = useStrategyCtx();
```
zavelevsky, line 48: in theory - this makes a single item sensitive to any change to the collection.
I guess that in practice, in our specific case, this is already the case - but I'd look for a way to avoid passing in the strategies collection just so that we can pass it to the event service

src/pages/explorer/index.tsx
```diff
@@ -26,3 +29,31 @@ export const ExplorerPage = () => {
     </Page>
   );
 };
+
+const ExplorerEvents = () => {
+  const [mounted, setMounted] = useState(false);
```
zavelevsky, line 34: probably no real reason to check if mounted

GrandSchtroumpf: You need it if for type change and filter change. When you arrive in the page you should only trigger search event. Type and filter should be triggered only when they change from one state to the next, the initial state doesn't count as the search already send it to analysis

src/components/strategies/overview/strategyBlock/StrategyBlockManage.tsx
GrandSchtroumpf: I'm not sure to understand. Strategies are wrapped in a context, so it's shared no matter how many components listen to it. We are using it in many places in the app already. Using it in this component should have not effect in theory.

PR #808, opened 2023-09-14
Improve performance on Explorer search

### LINKED ISSUE #803: [Performance] Explorer
Currently the Explorer has some performance issues, specifically on filtering strategies with the combobox.
I'll use this issue to track down what could cause the performance issue's sources and how to solve them

- [x] `filteredPairs`
- **Source**: `filteredPairs` is set in the the `ExplorerPage` but is only used in the `SuggestionList`
- **Solution**: Move filter & sorting of pairs into the `SuggestionList` to reduce rerendering

- [x] `search`
- **Source**: It's used for filtering suggestion, setting `routeParams` and for query data in the `ExplorerPage`
- **Solution**: use `search` for filtering suggestion, `routeParams` should come from the input.value on submit, use the routeParams to query data. This will also reduce the amount of rerendering.

- [x] `useExplorer`
- **Source**:  It is used in `ExplorerPage`, `ExplorerTypeOverviewPage`, `ExplorerTypePortfolioPage` & `ExplorerTypePortfolioTokenPage` which means it's triggered twice on the same page.
- **Solution**: Use a context in `ExplorerPage` to store the current strategies based on type & params. Use the context in the subpages.

GrandSchtroumpf:
fix #803 
- [x] move search & filterPair lower down the tree
- [x] use a context to store the pair maps
- [x] use `useExplorerParams` instead of passing type/slug on each component


pingustar (approved):
LGTM

PR #815, opened 2023-09-18
E2e/mock api

GrandSchtroumpf:
- [x] Create a script to fetch roi & market-rate (we shouldn't use it anymore except if we need to update the mock data for some reason)
- [x] Store the mock data in to the e2e folder
- [x] Create a utils to intercept request from the front toward carbon api and return mock data instead

Note: interceptor should be set on `beforeEach` and because the page is not yet available on `beforeAll`


scripts/mock-api.mjs
```diff
@@ -0,0 +1,178 @@
+import { writeFileSync, existsSync, mkdirSync } from 'fs';
```
zavelevsky, line 1: I'd add a comment here explaining what's this script is for

zavelevsky (approved):
LGTM

PR #816, opened 2023-09-19
Remove enum for string literal

### LINKED ISSUE #797: [Typescript] Switch from enum to string literal
Since string literals have been introduced in Typescript they are a better replacement for `enum`. 
Here is what enums look like once transpiled: 
image

Here is what string literals look like once transpiled: 
image

With string literals you still keep autocompletion:
image

And you can even extract them from JS values so you keep everything in sync : 
image

Forget old `enum` and join force with the string literals ! ⚔️🛡️

GrandSchtroumpf:
fix #797 
- [x] Strategy Status
- [x] MenuItemAction
- [x] ConnectionType
- [x] Provide migration for connection type


zavelevsky (approved):
Code LGTM

PR #833, opened 2023-09-25
[e2e] create strategy

### LINKED ISSUE #832: [e2e] Create Limit Strategy
- [x] navigate to My Strategies
- [x] click Create Strategy
- [x] pick two tokens using the token selection modal search
- [x] In the Set Prices page assert data correctness
- [x] fill numbers and create the strategy
- [x] assert that it appears with correct data in My Strategies 
- [x] assert that a notification appeared.

GrandSchtroumpf:
fix #832

Concerning the data correctness on the "Set Prices page" I currently check the 100% fill outcome, let me know if you see something else to assert.


e2e/pages/strategy.spec.ts
```diff
@@ -1,14 +1,80 @@
...
+import { test, expect } from '@playwright/test';
+import { navigateTo, screenshot, waitFor } from '../utils/operators';
 import { mockApi } from '../utils/mock-api';
+import { setupImposter } from '../utils/debug';
 
 test.describe('Strategies', () => {
-  test.beforeEach(({ page }) => {
-    mockApi(page);
+  test.beforeEach(async ({ page }) => {
+    await Promise.all([mockApi(page), setupImposter(page)]);
```
zavelevsky, line 8: why calling `mockApi(page)` twice?

e2e/pages/strategy.spec.ts
```diff
@@ -1,14 +1,80 @@
...
+    await waitFor(page, 'modal');
+    await page.getByLabel('Select Token').fill('dai');
+    await page.getByTestId('select-token-DAI').click();
+    await page.getByText('Next Step').click();
+
+    // Fill Buy fields
+    const buy = page.getByTestId('buy-section');
+    await buy.getByTestId('input-limit').fill('1500');
+    await buy.getByTestId('input-budget').fill('5000');
+    await expect(buy.getByTestId('outcome-target')).toHaveText('3.33 ETH');
```
zavelevsky, line 39: "target" is a confusing term here - as we use in the context of trade, and in strategies we talk base/quote (and not source/target).

e2e/pages/strategy.spec.ts
```diff
@@ -1,14 +1,80 @@
...
     await page.getByTestId('first-strategy').waitFor({ state: 'visible' });
     await screenshot(page, 'first-strategy');
   });
+
+  test('Create Limit Strategy ETH->DAI', async ({ page }) => {
+    await navigateTo(page, '/');
+    await page.getByTestId('create-strategy-desktop').click();
+
+    // Select Base
+    await page.getByTestId('select-base-token').click();
```
zavelevsky, line 22: Making these tests more data driven will make them reusable.
This test can be made more generic so that it can test any pair, and so you can take the following approach (pseudo code):
```js
const testCases = [
  {
    base: 'ETH',
    quote:  'DAI',
    buyLimit: 1500,
    buyBudget: 4000,
   .... //etc
  }
];

testCases.forEach(case => test(`Create Limit Strategy ${case.base}->${case.quote}`, ...
...

```

it will make it much easier to quickly add additional pairs and test cases

GrandSchtroumpf: I agree, but before jumping on a refactoring I would first do 2 more tests to see what is common or not.

e2e/pages/strategy.spec.ts
GrandSchtroumpf: It's the name used in the FullOutcome component.
On buy it's `base` and on sell it's `quote`. Is there specific semantic to describe this ?

e2e/pages/strategy.spec.ts
GrandSchtroumpf: Alright, let's build a utils for that

zavelevsky: ok, decided to refactor once patterns emerge

e2e/pages/strategy.spec.ts
zavelevsky: we go with `value`

src/components/strategies/create/BuySellBlock/BugetSection.tsx
```diff
@@ -30,6 +30,7 @@ export const BudgetSection: FC<Props> = ({
 }) => {
   const inputId = useId();
   const budgetToken = buy ? quote : base;
+  const testIdPrefix = buy ? 'buy' : 'sell';
```
tiagofilipenunes, line 33: testIdPrefix declared but not used

e2e/README.md
tiagofilipenunes, line 1: Got this error when running, perhaps add to readme for the user to do 'yarn playwright install' before running or add it to yarn e2e script

Error: browserType.launch: Executable doesn't exist at /Users/user/Library/Caches/ms-playwright/chromium-1076/chrome-mac/Chromium.app/Contents/MacOS/Chromium

tiagofilipenunes (changes requested):
Great work, phenomenal PR, very smooth testing experience! 

Comments: 
1. Should firefox and webkit be enabled in playwright.config.ts? I ran the tests using all 3 and they passed. Had to increase the timeout of the create strategy test slightly. 
Pasted image 20230928000114

2. The tenderly baseURL in e2e utils should not be hardcoded to avoid exposing the path, and to allow other contributors to run the tests locally without modifying the code if they don't have access to the tenderly project

GrandSchtroumpf:
@tiagofilipenunes Thanks for the review : 
1. Concerning Firefox & Safari we decided to use only chrome for now because we'll need to define a specific strategy for screenshot. If you think it's important, maybe we can open an issue to discuss that.

2. Concerning the tenderly path, there is nothing secret, as long as you don't have any api key they cannot do anything. Contributors will still need to change the code to remove setting up the fork & removing it, and test will fail if they don't open a fork at the right blockNumber anyway.

tiagofilipenunes (approved):
LGTM!

PR #851, opened 2023-10-05
Check input is '...' in the trade page

### LINKED ISSUE #850: Fix: Decimal input is sometimes '...'
Sometimes the input for Decimal() is '...', causing a DecimalError.

tiagofilipenunes:
fix #850


src/components/trade/tradeWidget/useBuySell.ts
```diff
@@ -173,7 +173,11 @@ export const useBuySell = ({
 
   useEffect(() => {
     if (byTargetQuery.data) {
-      if (new Decimal(targetInput).gt(new Decimal(liquidityQuery.data || 0))) {
+      if (
```
GrandSchtroumpf, line 176: It's not super readable, could you create variables with good names so it's easier to understand what is this condition about

src/components/trade/tradeWidget/useTradeAction.ts
```diff
@@ -49,8 +49,12 @@ export const useTradeAction = ({
 
   const calcMaxInput = useCallback(
     (amount: string) => {
+      if (amount === '...') return 'NaN';
```
GrandSchtroumpf, line 52: Why do you return NaN as a string ? It doesn't look like a good practice

src/components/trade/tradeWidget/useBuySell.ts
tiagofilipenunes: Made it clearer

zavelevsky: we can consider creating a utility function, something in the lines of (coding from memory)
```ts

function tryDecimal(value: number|string|Decimal) : Decimal | NaN {
  try {
    return new Decimal(value);
  } catch {
    return NaN;
  }
}

```

zavelevsky: and of course use it only when the input is questionable

tiagofilipenunes: Updated the implementation

PR #856, opened 2023-10-11
Fix popular search on trade

GrandSchtroumpf:
Since we use symbols to index the pairs, we need to use symbols for popular pairs too.


tiagofilipenunes (approved):
LGTM

PR #861, opened 2023-10-13
Set fix height to avoid layout shift

GrandSchtroumpf:
fix #856 
- [x] Set fix height to avoid layout shift
- [x] Improve a11y on trade page


zavelevsky (approved):
LGTM

PR #869, opened 2023-10-17
Fix routing modal

### LINKED ISSUE #865: Routing modal is broken
image

GrandSchtroumpf:
fix #865 

Before: 
image

After: 
image


ashachaf (approved):

zavelevsky (commented):
code LGTM but I would add min-height of 100px more or less to the routes element so that it won't disappear completely on short screens

zavelevsky (approved):
LGTM

PR #874, opened 2023-10-19
update query on selected currency change

### LINKED ISSUE #873: Update market rate after currency update
https://github.com/bancorprotocol/carbon-app/assets/8143464/8027759c-fd15-4c61-ba71-789703874197

GrandSchtroumpf:
fix #873

Because the server wasn't able to provide the fiat price for multiple currency we had to change the query for one currency only.
So now we have to update query on currency change


tiagofilipenunes (approved):
LGTM as a temporary fix. Note: useGetMultipleTokenPrices no longer provides multiple token prices.

ashachaf (approved):

PR #891, opened 2023-10-26
Expose routing option for every valid amount

### LINKED ISSUE #887: Trade - expose routing option for every valid amount
as a user i would like to have the option to view the routing modal for any input amount which is valid, even if it is greater than my wallet balance.

### insufficient budget error
this flow checks if the user input amount =< wallet balance for the token. 
in this flow, we should do the following:
1. disable the CTA (or make it do nothing)
2. display "routing" option. if the user uncheck orders, we should re-validate input amount =< wallet balance and if true, enable the CTA to trigger the transaction.

references:
in this image, the insufficient balance error is visible. we should expose the "routing" option
image

in this image, i was able to access routing with input amount > wallet balance.
- CTA should be disabled
- we should display "insufficient balance" error
image

in this image, i have only left 1 order creating a situation where input amount < wallet balance
image

GrandSchtroumpf:
fix #887 

- [x] disable both CTA if insufficient balance
- [x] display routing button if insufficient balance

Bonus: 
- [x] Add a check all button to quickly select one
- [x] Support keyboard navigation (`Ctrl + a`, `ArrowDown`, `ArrowUp`, `Home`, `End`)

**Disabled** :
image

**Enabled**:
image


zavelevsky (approved):
code LGTM

ashachaf (approved):

PR #893, opened 2023-10-27
Explorer search border & CTA button focus

### LINKED ISSUE #883: Keyboard navigation stuck in Firefox
When navigating with keyboard, the Tab get stuck on the "Create Strategy" button. I think this is because it's a link inside a button, we should remove the button around.
https://github.com/bancorprotocol/carbon-app/assets/8143464/f5c51c9e-46c2-4852-9f4e-2965f296062b

GrandSchtroumpf:
This PR tackles two small UI issues

## Explorer Search border
The border should become green when the user focuses the search providing better feedback and reducing noise when he's not interacting with the search.

Idle:
image

Focus: 
image

Focus with error:
image

Idle with error:
image

## Fix focus trap
fix #883
Use Link instead of link inside a button to avoid focus being stuck on the CTA


zavelevsky (approved):
code LGTM

ashachaf:
is it possible that the search bar will have "focus" design on page load?
flow:
1. user navigate to /explorer
2. search bar state = focus
3. user input = focus
4. user out of focus = ideal

PR #901, opened 2023-11-02
Fix keyboard navigation on firefox

### LINKED ISSUE #883: Keyboard navigation stuck in Firefox

GrandSchtroumpf:
fix #883 
- [x] Remove button around link


zavelevsky (approved):
LGTM

PR #915, opened 2023-11-16
Fallback with simpler regex if emoji not available

### LINKED ISSUE #914: Explorer search by wallet is broken on older browsers
Go to an old chrome version, visit the explorer page

1. enter some invalid text as a wallet address. Expected behavior is to see an indication of invalid input - but you don't see it
2. Enter a valid wallet address and click search - expected behavior is for the content of the page to change in order to either display strategies of an indication that none was found. What actually happens is nothing - as if the button wasn't clicked.
3. Navigate directly to a specific wallet in the explorer. Expected behavior - see the wallets strategies. Actual behavior - it's stuck loading.

GrandSchtroumpf:
fix #914 
We were looking for emojis in the ENS address. It's only available since chrome 112.
If the regex fails, we fallback to a simpler regex.

In any case we do a request to the ENS resolver to verify if address exist.


zavelevsky (approved):
LGTM

PR #917, opened 2023-11-17
Fix maximumFactionDigits & currency display on old browser

### LINKED ISSUE #908: Intl NumberFormat
Some old browser throw `maximumFractionDigits out of range`.
We need to understand why and either fix it or use a polyfill for these browsers only.
Currently we use the polyfill all the time as a hotfix

GrandSchtroumpf:
fix #908 
- [x] set minimumFactionDigits to 0 for number gt 100.
- [x] remove polyfill as it work on chrome 80
- [x] use symbol display by default


pingustar (approved):
Code LGTM

PR #921, opened 2023-11-20
Support reload on edit strategy

### LINKED ISSUE #459: Edit Strategy - Add deeplink to withdraw, deposit, renew, edit prices
When user edit a strategy, we want to allow the user to refresh the page and remain at the same page and state.
Currently, if user refresh the page in edit strategy page, he will redirect to strategy overview page.

we have already have a pr opened for that one that need to be polished - https://github.com/bancorprotocol/carbon-app/pull/458
branch - 421-edit-strategy-page-deeplink

Also fix these issues: 
- [ ] Split useUpdateStrategy hook. Create custom hook for each edit strategy.
- [ ] in ModalConfirmStrategy.tsx, 
    const isAwaiting =
    type === 'delete' ? deleteMutation.isLoading : updateMutation.isLoading;
    
    need to think of better idea of decoupling those edit features.

GrandSchtroumpf:
fix #459 
- [x] use strategy id in the edit page
- [x] remove strategyToEdit from the store


src/libs/routing/pathnames.ts
```diff
@@ -21,7 +21,7 @@ export const PathNames = {
   trade: '/trade',
   debug: '/debug',
   createStrategy: '/strategies/create',
-  editStrategy: '/strategies/edit',
+  editStrategy: '/strategies/edit/$strategyId',
```
pingustar, line 24: interesting, this must be new in the latest router update. We should consider to apply this new way of handling params to the explorer as well.

just a note to myself, no action required.

src/pages/strategies/edit/index.tsx
```diff
@@ -1,25 +1,30 @@
...
+      if (strategy) setStrategy(strategy);
+      else navigate({ to: PathNames.strategies });
     }
-  }, [user, strategyToEdit, navigate]);
+  }, [user, strategyId, strategies, isLoading, navigate]);
 
-  return user && strategyToEdit ? (
-    <EditStrategyMain strategy={strategyToEdit} />
+  return user && strategy ? (
+    <EditStrategyMain strategy={strategy} />
```
pingustar, line 27: I think the loading state should be improved. Currently it shows the overview page. We should handle this a bit more elegant and show the loading animation.

And show a error message if the id is not found.

src/pages/strategies/edit/index.tsx
```diff
@@ -1,25 +1,30 @@
...
+  const [strategy, setStrategy] = useState<Strategy>();
   const navigate = useNavigate();
 
   useEffect(() => {
-    if (!user || !strategyToEdit) {
+    if (isLoading) return;
+    if (!user || !strategyId) {
       navigate({ to: PathNames.strategies });
+    } else {
+      const strategy = strategies?.find(({ id }) => id === strategyId);
```
pingustar, line 20: we might wanna use the "displayId" for the url parameter. We currently dont expose the actual contract id to the user so it would be more userfriendly and consistent with the ui. But thats a question for @ashachaf

pingustar (commented):
Code LGTM. Left some comments for improvements

src/pages/strategies/edit/index.tsx
GrandSchtroumpf: I agree, but it would require new mockup for this state. This implementation match the current behavior (fallback to overview page), so I think we can address the loading/error state in another PR. What do you think @ashachaf ?

src/libs/routing/pathnames.ts
zavelevsky: how is this different from /:strategyId ?

zavelevsky (approved):
LGTM

src/libs/routing/pathnames.ts
GrandSchtroumpf: Nothing different, just the new syntax I guess: 
https://tanstack.com/router/v1/docs/guide/path-params

pingustar: > how is this different from /:strategyId ?

we use the pathnames in the app for navigation with `Link` or `useNavigation`. The old router was not able to replace `:strategyId` or `$strategyId` with the expected variable without passing it as a function argument.

```ts
{
  ...
  explorer: (type: ExplorerType) => `/explorer/${type}`,
  ...
}
```

the new version can handle this now

```ts
<Link to="/blog/$postId" params={{ postId: '123' }}>
      Post 123
 </Link>
```

instead of

```ts
<Link to={Pathnames.blog(postId)}>
      Post 123
 </Link>
```

ashachaf:
- [ ] can/should we add the loading animation?
https://github.com/bancorprotocol/carbon-app/assets/40430285/c82dc7be-6daf-4bb8-894e-fb2c7b2fdc29

src/libs/routing/pathnames.ts
GrandSchtroumpf: https://github.com/bancorprotocol/carbon-app/issues/969

ashachaf (approved):

PR #924, opened 2023-11-24
Duplicate strategy has the correct create UI

### LINKED ISSUE #760: Bug: Duplicating strategy results in undefined strategyType
When a user selects “Duplicate Strategy”, the strategyType field is undefined, leading to erroneous information and potentially behaviour too.

How to replicate: Press “Duplicate Strategy” from any strategy modal. The type of strategy defaults to Disposable even if it isn’t, with matching tooltips. 

Example: Disposable strategy text when the strategy is recurring. Tooltips are for disposable strategies too.

telegram-cloud-photo-size-4-6014926768781245483-y

### LINKED ISSUE #923: Remove Buffer in favor of btoa & atob
We are using `Buffer` to convert the strategy into base64 when we duplicate it.
Buffer is a Node.JS module. The browser supports `atob` & `btoa` to transform string to base64, let's use that instead to avoid adding more bundle size.

GrandSchtroumpf:
fix #923
fix #760

- [x] use atob & btoa instead of Buffer to encode/decode base64
- [x] derive type, setting & direction from decoded strategy


zavelevsky (approved):
LGTM

ashachaf:
- [ ] when duplicating limit order, we cannot carry over the budget for the inactive order
flow:
rpc: https://rpc.tenderly.co/fork/04b217d4-a982-4b80-b8ff-e07be42c2bc4
https://issue--760.carbon-app-csq.pages.dev/explorer/token-pair/link_eth
1. duplicate strategy 692 
2. it opens a "buy" disposable interface (which is correct)
3. however, it forces the user to deposit the sell budget (from the hidden order with prices of 0)
expected:
open the buy screen with prices and budget, clear the "sell" budget input.

ashachaf (approved):

Issue #936, opened 2023-12-14
strategy RESET/MAINTAIN 


ashachaf:
- [x] edit limit order
when editing prices of a limit order, we reset z value for the order with no budget (or prices).
unlike creation where we calculate z value.
not a bug. just inconsistence behavior.

- [x] deposit/withdraw limit order
when deposit/withdraw of a limit order, we reset z value for the order with no budget (or prices).
unlike creation where we calculate z value.
we do not pass RESET/MAINTAIN, i only see 0.

- [x] renew limit order
when i reactivate a paused limit order, the order with no price or balance, receive z as 0.
unlike creation where we calculate z.
not a bug. just inconsistence behavior.

- [x] deposit funds recurring range
when depositing into an existing strat, 
1. we pass to the contract/sdk maintain/reset value for both budgets even if no balance is added (IF user input budget is null, omit this value)
2. we use RESET as a default value which result in resetting the order that the user didnt explicitly indicate to MAINTAIN
https://github.com/bancorprotocol/carbon-app/assets/40430285/3fb8364a-fb37-41aa-996a-62a596cbe191

- [x] create recurring ranges with 1 budget (i believe this is the same issue as above. we RESET the wrong order incorrectly)
<img width="389" alt="image" src="https://github.com/bancorprotocol/carbon-app/assets/40430285/c6f26607-ad2c-4eac-b58d-48798ca577b4">
<img width="687" alt="image" src="https://github.com/bancorprotocol/carbon-app/assets/40430285/7a5ba930-ad1c-4ad1-a828-08c2f1b5540c">

now, deposit 1 BNT into the BNT side (which was empty before)
<img width="416" alt="image" src="https://github.com/bancorprotocol/carbon-app/assets/40430285/1f4507f4-a9f2-44f4-8990-893f61c7ddaa">
<img width="686" alt="image" src="https://github.com/bancorprotocol/carbon-app/assets/40430285/5bb5d180-0853-4ba5-95a6-ca4a3992f9fe">
<img width="395" alt="image" src="https://github.com/bancorprotocol/carbon-app/assets/40430285/158fd2a3-cf70-4e31-bd3e-bc1a7f5e7bfb">

- [x] withdraw from recurring range strategy
we RESET the budget IF the user does not explicitly indicate to MAINTAIN.
even if the user did not withdraw any amount from that budget.

https://github.com/bancorprotocol/carbon-app/assets/40430285/43839341-2895-4527-833b-e41e4366e469

PR #939, opened 2023-12-15
#936 - Fix Strategy RESET/MAINTAIN

pingustar:
#936


ashachaf:
- [ ] deposit range strategy
1. create a strategy with both budgets
2. trade out everything from 1 order
3. deposit into the side with budget only
expect: the empty order should keep the same marginal price
actual: the marginal price gets reset somehow
(i believe we should NOT pass any value if there is no change. for this, we should NOT pass 0 budget for the empty order)
image

src/components/strategies/edit/EditStrategyBudgetContent.tsx
```diff
@@ -129,36 +130,38 @@ export const EditStrategyBudgetContent = ({
...
-        )
-      : depositBudget(
-          updatedStrategy,
-          order0.marginalPriceOption,
-          order1.marginalPriceOption,
-          handleEvents
-        );
+    const action = type === 'withdraw' ? withdrawBudget : depositBudget;
+
+    // TODO fix type error AtLeastOneOf fieldsToUpdate
```
zavelevsky, line 162: I thought you worked around that

ashachaf (approved):

src/components/strategies/edit/EditStrategyBudgetContent.tsx
pingustar: fixed

PR #953, opened 2023-12-23
wrong check in edit price of limit orders

zavelevsky:
The wrong check led to never being able to edit prices for limit strategy


pingustar (approved):
LGTM

PR #959, opened 2024-01-02
Prevent unnecessary calls to the SDK

GrandSchtroumpf:
- Do not call SDK if buyMin or sellMax is not set
- Do not call SDK if buyMin is 0, if buyMin == sellMax, spread == 0 or spread == 100


zavelevsky (approved):
LGTM

PR #965, opened 2024-01-03
Fixes on number input behavior

### LINKED ISSUE #963: [Create Disposable Limit] CTA enabled with 0
CTA is enabled if value is 0 in a disable limit strategy
image

In `src/components/strategies/create/useCreateStrategy.tsx` we've got :
```typescript
const isOrder0Valid = order0.isRange
  ? isValidRange(order0.min, order0.max)
  : +order0.price >= 0 && order0.price !== ''; // <--- Here
const isOrder1Valid = order1.isRange
  ? isValidRange(order1.min, order1.max)
  : +order1.price >= 0 && order1.price !== ''; // <--- Here
return !isOrder0Valid || !isOrder1Valid;
```

Because disposable limit order is either buy or sell, one of the two is going to be 0. We need to identify which one is selected before checking price

### LINKED ISSUE #964: [Create Strategy] Manage input with "."
On change value "." should be considered as "0"
On blur update the value as follow : 
- if value is "." -> "0"
- If value is ".12" -> "0.12"
- If value is "12." -> "12"

GrandSchtroumpf:
fix #964
fix #963
blocked by #924

- [x] Input with "." should be consider value 0
- [x] Sanitize input on blur (".2300" -> "0.23")
- [x] Do not call SDK if Number(value) is NaN
- [x] Rename spreadPPM to spread
- [x] Spread behaves like existing input
- [x] Remove trailing 0 when automatically setting min & max price 
- [x] Price 0 should disable CTA in disposable limit strategy
- [x] Reset anchor when market price changes and update marginal price to be equal to price


src/utils/helpers/index.ts
```diff
@@ -34,6 +34,11 @@ export const sanitizeNumberInput = (
   else return sanitized;
 };
 
+export const sanitizeInputOnBlur = (value: string) => {
```
zavelevsky, line 37: 1. Why not use `sanitizeNumberInput` (and also handle "." there)?
2. I'd rename this function. It's in utils should it shouldn't "dictate" when it's being used.

GrandSchtroumpf: 1. `sanitizeNumberInput` is used on change. "." should only be transformed to "0" only after the user focus out, else user won't be able to write `.23`.
`Number(".") === NaN` so I need a specific check for it instead of always do `Number(value).toString()`.
2. The function is (and should be) tightly link with the input cycle check (so is `sanitizeNumberInput`). Now that you mention it, I think the best would be to rename `sanitizeNumberInput` to `sanitizeNumberInputOnChange`. I can move these functions in a dedicated "input-number.ts" file.

GrandSchtroumpf: Could only be an object as they are supposed to be used in combination : 
```typescript
export const sanitizeNumberInput = {
  onChange: () => ,
  onBlur: () => ,
}
```

zavelevsky: sounds like a plan

src/components/strategies/create/overlapping/CreateOverlappingStrategy.tsx
```diff
@@ -40,10 +42,14 @@ export type SetOverlappingParams = (
   max: string
 ) => Promise<StrategyPrices | undefined>;
 
+const roundDecimal = (value: number, precision: number) => {
```
zavelevsky, line 45: I am pretty sure we already have such a util

Costa769 (approved):

ashachaf:
- [ ] spread input field behavior, should highlight the entire number on focus
https://github.com/bancorprotocol/carbon-app/assets/40430285/b11d29d3-08de-45dc-903f-bcedf7767f1a

zavelevsky (approved):
LGTM

ashachaf (approved):

GrandSchtroumpf:
close in favor of https://github.com/bancorprotocol/carbon-app/pull/980

PR #973, opened 2024-01-09
[E2E] create overlapping strategy

GrandSchtroumpf:
- [x] Fix existing test
- [x] Add screenshot
- [x] Add assertion on budget caculation


src/components/strategies/create/overlapping/CreateOverlappingStrategy.tsx
```diff
@@ -59,18 +59,18 @@ export const CreateOverlappingStrategy: FC<OverlappingStrategyProps> = (
   });
 
   const setOverlappingParams = async (min: string, max: string) => {
+    order0.setMin(min);
```
zavelevsky, line 62: What's the reason for a behavior change as part of a PR for tests? Is there any functional significance to moving these sets to the top?

GrandSchtroumpf: The E2E were too quick to fill min & max. So we had a concurrency issue with the initial input setup and the E2E writing into the input.
By not waiting for the SDK to give use the rest of the fields we prevent this race. The only impact I saw is that now update is done in 2 frames instead of 1.

zavelevsky (approved):
LGTM!!

GrandSchtroumpf:
close in favor https://github.com/bancorprotocol/carbon-app/pull/979

PR #974, opened 2024-01-09
Set mono font for address when in E2E

GrandSchtroumpf:
The screenshot are always different because the width of the address button always changes
By using a mono font when in the E2E test environment we prevent the width to change depending on the address.

- [x] Create a debug field to set environment as E2E
- [x] Change font of the address to mono when in E2E
- [x] Update all `beforeEach` to set the E2E environment


zavelevsky (approved):
LGTM

PR #978, opened 2024-01-10
E2e-tests-restructure

tiagofilipenunes:


e2e/utils/strategy/template.test.ts
```diff
@@ -2,9 +2,9 @@ import { describe, test, expect } from 'vitest';
 import { createDebugStrategy } from './template';
 
 describe('Test create strategy template', () => {
-  // Limit
-  test('limitBuy', () => {
-    const strategy = createDebugStrategy.limitBuy('ETH->DAI', {
+  // Recurring Limit
+  test('recurringLimitBuy', () => {
```
GrandSchtroumpf, line 6: I think you shouldn't change these tests. The mapping is like that :
- limitBuy -> disposable limit buy
- limitBuySell -> recurring limit_limit

The goal of these function was to build easily a testCase by doing : 
```
[
  limitBuy('ETH->DAI', { price: '100', budget: '100' }),
  ...
]
```
But it might have been premature work, so let's keep it as it was before.

e2e/utils/strategy/template.ts
```diff
@@ -61,17 +65,23 @@ const fromLimitOrder = (order: LimitOrder): RangeOrder => ({
   budget: order.budget,
 });
 export const createDebugStrategy = {
-  limitBuy: (pair: TokenPair, buy: LimitOrder): CreateStrategyTemplate => ({
+  recurringLimitBuy: (
```
GrandSchtroumpf, line 68: same comment as above. I would leave this file as it for now, we can improve it later when unifying the naming

e2e/utils/strategy/template.ts
```diff
@@ -27,7 +27,11 @@ interface LimitOrder {
   budget: string;
 }
 
-export const STRATEGY_TYPES = ['Limit', 'Range', 'Overlapping'] as const;
+export const STRATEGY_TYPES = [
+  'Recurring',
```
GrandSchtroumpf, line 31: I think it should be lowercase like the rest.

GrandSchtroumpf (approved):
Very nice work. I think the overlapping & disposable folders should only have what they export. Like that it'll be easier to review for next PR as it'll be a file creation and not an update.
```
disposable/
  delete.ts
  index.ts
overlapping/
  create.ts
  index.ts
```

PR #980, opened 2024-01-11
Fix input number behavior

### LINKED ISSUE #963: [Create Disposable Limit] CTA enabled with 0

### LINKED ISSUE #964: [Create Strategy] Manage input with "."

GrandSchtroumpf:
fix #964
fix #963
blocked by #924

- [x] Input with "." should be consider value 0
- [x] Sanitize input on blur (".2300" -> "0.23")
- [x] Do not call SDK if Number(value) is NaN
- [x] Rename spreadPPM to spread
- [x] Spread behaves like existing input
- [x] Remove trailing 0 when automatically setting min & max price 
- [x] Price 0 should disable CTA in disposable limit strategy
- [x] Reset anchor when market price changes and update marginal price to be equal to price


zavelevsky (approved):
LGTM

PR #983, opened 2024-01-11
Small number display - subscript

### LINKED ISSUE #929: Number display - small number handling
This issue is taking a piece out of issue #727.

# Overview
we would like to update the way small numbers are displayed. 
currently, we show 6 decimals and if the number is smaller we show "<0.000001"
the updated approach would expose the small numbers to the user in an updated manner.

### Exceptions

- Percentage numbers ->display logic for percentage numbers remains unchanged, and percentages below 0.01% will continue to be displayed as they are currently, i.e., "<0.01%."
- Input fields->both user input and automatically prefilled values will continue to  display full numbers without any changes.
- When mouse over a subscribe number and the tooltip is displayed, display full number in the tooltip without applying the subscribe format
- 
## Mockup

[Designs](https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=11296%3A86130&mode=design&t=nxpkUEwRyQ6TTWyU-1)

## Specifications
**IF** {integer digits}=0 AND {number of first decimals that are 0}>=3 do the following:
- Display first decimal as 0
- Count the number of 0 decimals from the first decimal up to the first decimal that is not 0
- Display the number as a subscript next to the first 0 decimal (if counted 6 zeros, the number needs to be 6 and the first zero is visible)
- Take 5 decimals starting from the first decimal that is not 0 and display them after the first 0 decimal
image

**Examples(prices displayed in USD):**
image

GrandSchtroumpf:
fix #929 

- [x] Use subscript for number, price but not percent
- [x] Display full value for tooltips **except** in the strategy card's graph
- [x] Refactor how digits are displayed for price
- [x] Update unit tests for `prettifyNumber`
- [x] Update e2e with new values

## Digits for price : 
```
111 -> 111.00
11 -> 11.00
1 -> 1.00
0 -> 0.00
0.111 -> 0.11
0.0111 -> 0.011
0.00111 -> 0.0011
0.000111 -> 0.0₃111
```

## My Strategies
image
image
image
image
image
image

## Trade
image
image
image
image

## Explorer
image
image

## Create Strategy
image
image
image
image

## Edit Strategy
image
image
image


e2e/screenshots/[Create Overlapping Strategy] Form.jpg
Comment (zavelevsky) [inline e2e/screenshots/[Create Overlapping Strategy] Form.jpg:1]: what's the correct state of the CTA

Comment (GrandSchtroumpf) [reply e2e/screenshots/[Create Overlapping Strategy] Form.jpg:1]: After every input has been filled the CTA should be enable

zavelevsky:
image

the price display shows the subscript numbers to high relative to the line

zavelevsky:
image

the digits are too spaced apart - it looks better in other places where it was implemented

zavelevsky:
image

here the digits are too down - their center should be on the bottom of the row

src/utils/helpers/number.test.ts
```diff
@@ -32,195 +32,220 @@ describe('Test helpers', () => {
...
-      expect(prettifyNumber(0.0000001)).toEqual('< 0.000001');
-    });
+      test('should return "0.0₅1" for input 0.000001 and no currentCurrency selected', () => {
+        expect(prettifyNumber(0.000001)).toEqual('0.0₅1');
+      });
 
-    test('should return "1,321,965,595" for large number', () => {
-      expect(prettifyNumber(1321965595)).toEqual('1,321,965,595');
-    });
+      test('should return "0.0₆12345" for input 0.000000123456 and no currentCurrency selected', () => {
```
zavelevsky, line 204: I don't see tests for negative exponents (e-4), a case you are covering in the code

GrandSchtroumpf: All small number in the unit test are transformed into e-x by the SafeDecimal.toString().
I actually had to add this case because of unit test.
I think I can add test for the `subscript` function itself

zavelevsky:
@ashachaf @Costa769 the current font dictates the behavior of the subscript digits too - so monospace means they are aligned with the bottom of the line and are evenly spaced, and the regular font dictates they are completely under the line. 
We can change fonts. We can change expectations. We can force behavior that's no the font's one by using components

Costa769:
Depth chart
 

1. Price is not displayed in the correct format

image

2. Incorrect font
image

ashachaf:
- [ ] orderbook missing display currency 
in the preview link:
image

production:
image

- [ ] orderbook incorrect prices
strategy prices:
image

orderbook prices: (need to check if we trim the data the frontend receives)
image

- [ ] same as prod so should be handled in a different issue
image

- [ ] aim to keep in the same line 
image

- [ ] spacing (not an issue, just FYI)
GT America Mono has different spacing which makes the numbers and gap between words a bit larger than GT America
image

Issue #951, opened 2023-12-20
Trade Page - Pair modal loading animation


pingustar:
Token Pair modal loading animation is missing

https://github.com/bancorprotocol/carbon-app/assets/7262066/0455bc9e-6567-4c54-a975-830c85fca099

PR #986, opened 2024-01-15
#951 - Fix modal styling

pingustar:
#951


GrandSchtroumpf (approved):
Code LGTM. So the issue was to use placeholderData instead of initialData ?

pingustar:
> Code LGTM. So the issue was to use placeholderData instead of initialData ?

the main problem with the loading animation was that the modal background was changed to the same color as the animation skeleton - the both are `silver` causing the loading animation to be invisible.

I fixed it by changing the loading animation to be black color.

Regarding the change of placeholderData instead of initialData, that is something I noticed while going through the code. 

Initial Data is the more suitable feature for our usecase.

A nice writeup about the difference you can find here: https://tkdodo.eu/blog/placeholder-and-initial-data-in-react-query

Issue #987, opened 2024-01-16
Price chart doesn't show on the set prices page of the overlapping strategy


Costa769:
Price range section doesn't show the price chart on set prices page of the overlapping strategy despite having the market price data of the base token 

## Flow

1. Click on create strategy CTA on my strategies page
2. Select PEPE/USDT token pair
3. Select overlapping strategy
4. Click on "Next step" CTA

### Excepted result

On "Set prices" page load
1. Show a chart in the Price range section
2. Trading view price chart is expended

### Actual result
On "Set prices" page load
1. Price range section is empty
2. Trading view price chart is collapsed

If user enters the prices in the set prices section , the chart appears and shows the market price of the base token

Issue #972, opened 2024-01-08
[Overlapping] Graph possible range is limited


GrandSchtroumpf:
If market price is ~1000 and price range is below 0.00001, the graph is broken
![image](https://github.com/bancorprotocol/carbon-app/assets/8143464/71b45cb2-c99f-4012-85e2-c2e1a2f3c314)

GrandSchtroumpf:
This happens when there is a difference of  6 digits between the market price and the price.
For example: If market price is 1000, then you see it with 0.0001 or 1_000_000_000

PR #989, opened 2024-01-16
Overlapping graph with extreme values

GrandSchtroumpf:
fix https://github.com/bancorprotocol/carbon-app/issues/987
fix https://github.com/bancorprotocol/carbon-app/issues/972

- Initialize min & max differently if market price is below 1wei
- Change graph ratio to support large difference between price range and between market price & price.
- Ensure price initialization in production mode


src/components/strategies/create/overlapping/CreateOverlappingStrategy.tsx
```diff
@@ -121,10 +130,14 @@ export const CreateOverlappingStrategy: FC<OverlappingStrategyProps> = (
   useEffect(() => {
     if (!quote || !base || marketPrice <= 0) return;
     if (!order0.min && !order1.max) {
-      const _marketPrice = new SafeDecimal(marketPrice);
-      const min = _marketPrice.times(0.999).toFixed(quote.decimals);
-      const max = _marketPrice.times(1.001).toFixed(quote.decimals);
-      if (isValidRange(min, max)) setOverlappingParams(min, max);
+      const { min, max } = getInitialPrice(new SafeDecimal(marketPrice), quote);
```
pingustar, line 133: I would move `new SafeDecimal(marketPrice)` into `getInitialPrice` and accept `number | string` as marketPrice argument type.

src/components/strategies/create/overlapping/CreateOverlappingStrategy.tsx
```diff
@@ -44,6 +44,15 @@ export type SetOverlappingParams = (
   max: string
 ) => Promise<StrategyPrices | undefined>;
 
+const getInitialPrice = (marketPrice: SafeDecimal, quote: Token) => {
```
pingustar, line 47: is this only for quote? Might be worth to rename `quote` to `tkn` and make it a bit more universal.

pingustar (approved):
Code LGTM. Left some nit comments

src/components/strategies/create/overlapping/CreateOverlappingStrategy.tsx
GrandSchtroumpf: All price are display in quote in the app so it'll always be quote

ashachaf:
- [ ] when dragging the handles in the chart, the input prices do not change to reflect
- [ ] at some point, when dragging the handles, the price reset to a high value for some reason
https://github.com/bancorprotocol/carbon-app/assets/40430285/0989ddde-d4a3-483c-8284-4642ff42d1df

PR #990, opened 2024-01-16
[E2E] use addresses instead of symbol for token selection

GrandSchtroumpf:
Use addresses instead of symbol for token selection.
We currently limit the tokens to the list in debug. This can be extended later.


zavelevsky (approved):
LGTM

PR #1006, opened 2024-01-26
Chart works with values below 1wei

### LINKED ISSUE #988: [Overlapping] Chart not working well on a 1wei scale
PEPE -> USDC creates 2 issues with price on overlapping strategy because it behaves on a 1wei scale

## Market price a 1wei
We cannot apply +/- 0.1% as initial price for PEPE -> USDC has price goes below 1 decimal.
The current workaround is to set :
- min: 1wei
- max: 10wei

Why 10wei ? Because graph currently doesn't work well on a 1 wei scale (see below), so we need a larger delta between min & max

## Graph is not working correctly on a 1wei scale
- text is not correct
- polygons get stuck in between two weis
- when price get too close the graph breaks and automation price kicks in

https://github.com/bancorprotocol/carbon-app/assets/8143464/a5a73abb-4c5b-4bca-b84f-7da2b38d6615

GrandSchtroumpf:
fix #988 

- [x] update graph to support number below 1wei
- [x] update SDK version to have sync function & full price decimal
- [x] cleanup code after sync function
- [x] verify above/below market on every change
- [x] Remove unused utils


e2e/pages/strategy.spec.ts
```diff
@@ -520,7 +520,7 @@ const testCases: CreateStrategyTestCase[] = [
       create: {
         buy: {
           min: '0.3',
-          max: '0.545454',
```
zavelevsky, line 523: why would the numbers change?

GrandSchtroumpf: Because SDK doesn't use `toFixed` anymore.

e2e/screenshots/strategy/disposable/Disposable buy range/create/my-strategy.jpg
zavelevsky, line 1: How is the diff in the range explained?

e2e/screenshots/strategy/recurring/Recurring limit range/deposit/form.jpg
zavelevsky, line 1: tooltip still showing

e2e/screenshots/strategy/recurring/Recurring range limit/deposit/form.jpg
zavelevsky, line 1: tooltip still showing

ashachaf (approved):

zavelevsky (approved):
LGTM

PR #1010, opened 2024-01-29
using sync functions of the SDK for overlapping flows

zavelevsky:
This PR makes use of the latest version of the SDK in which breaking change was introduced to calculate overlapping prices - namely not requiring the quote token address and not limiting prices to a 1 wei resolution.

Switched over the app to use sync functions instead of async - given that the app has the token decimals to provide.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "private": true,
   "dependencies": {
     "@babel/core": "^7.0.0-0",
-    "@bancor/carbon-sdk": "^0.0.94-DEV",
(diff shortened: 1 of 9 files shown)


GrandSchtroumpf (approved):
❤️

GrandSchtroumpf (changes requested):
Now when the SDK throws, it breaks the app : 
image

zavelevsky:
handled as part of another PR

PR #1012, opened 2024-01-29
Simulator token and type selector

tiagofilipenunes:


src/components/simulator/SimulatorStrategyType.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+interface Props {
+  strategyType: SimulatorType;
+}
+
+export const SimulatorStrategyType: FC<Props> = ({ strategyType }) => {
+  const list = useRef<HTMLUListElement>(null);
+  const navigate = useNavigate();
+
+  const items: {
```
pingustar, line 17: lets declare a local interface for readability

src/components/simulator/SimulatorStrategyType.tsx
```diff
@@ -0,0 +1,105 @@
...
+      </header>
+      <ul
+        onKeyDown={handleKeyDown}
+        ref={list}
+        role="tablist"
+        className={`grid grid-cols-2 gap-8`}
+      >
+        {items.map(({ label, svg, tooltipText }) => (
+          <li role="none" key={label} className="relative">
+            <button
```
pingustar, line 65: I would probably use a <Link> component here. Reason is, that its just a simple navigation and you benefit from stuff like right-click it and open in new tab.

pingustar (approved):
nice work!

src/components/simulator/SimulatorStrategyType.tsx
tiagofilipenunes: done

src/components/simulator/SimulatorStrategyType.tsx
tiagofilipenunes: done

PR #1025, opened 2024-02-02
Add small budget warning

### LINKED ISSUE #1016: [Overlapping] Budget should be above other budget wei
Add a warning and disable CTA if the one budget is lower than 1wei.
In this example 1PEPE returns a budget of USDC lower than 1wei USDC, so it returns 0.
image

How to verify : 
if both budget are enable and one of them is 0, but not the other.
```typescript
export function isBudgetBelowWei(order0, order1) {
  if (isMaxBelowMarket(order1)) return false;
  if (isMinAboveMarket(order0)) return false;
  if (order0.budget && order1.budget) return false;
  if (!!order0.budget && !order1.budget) return false;
  return true;
}
```

GrandSchtroumpf:
fix #1016

- [x] Add warning when budget is too small

Warning on create
image

Warning on withdraw
image

Warning on deposit
image


ashachaf:
- [ ] trim the warning message
current: 
Calculated budget for USDC is smaller than its decimals. Please increase PEPE budget to have at least 0.0₆1 USDC
updated:
Please increase PEPE budget to have at least 0.0₆1 USDC

- [ ] above/below market price
in cases where the strategy is above/below market price and the user can only indicate 1 budget (by design), we need to apply the same process.
```IF calculated yint < 1 wei```
currently when strategy is above market price
image

expected behavior
image

File: src/components/strategies/create/overlapping/CreateOverlappingStrategyBudget.tsx [hunk 1]
```diff
@@ -1,12 +1,17 @@
...
 import { OrderCreate } from '../useOrder';
 import { ReactComponent as IconLink } from 'assets/icons/link.svg';
 import { OverlappingStrategyProps } from './CreateOverlappingStrategy';
 import { SafeDecimal } from 'libs/safedecimal';
-import { isMinAboveMarket, isMaxBelowMarket } from '../../overlapping/utils';
+import {
+  isMinAboveMarket,
+  isMaxBelowMarket,
+  isOverlappingBudgetTooSmall,
+} from '../../overlapping/utils';
```
zavelevsky, line 11: relative path

ashachaf (approved):

PR #1037, opened 2024-02-09
Centralized config files

### LINKED ISSUE #1024: It takes changes in too many places to get the app to work on another network
In order to run the app with another network the developer should:
- set the RPC under .env - VITE_CHAIN_RPC_URL 
- modifying the files under  /src/services/web3 to reflect the correct tokens and addresses
- once the backend is setup - its URL should be set in src/utils/carbonApi.ts

We should ideally have a single place for all these settings

GrandSchtroumpf:
fix #1024 

Create a config folder to manage variables related the network : 
```
config/
  [network]/
    [mode].ts
```
with : 
- network: `'ethereum'`
- mode: `'development' | 'production'`


src/config/mainnet/production.ts
zavelevsky, line 1: should it be mainnet or ethereum? I think the latter

src/config/mainnet/common.ts
zavelevsky, line 1: mainnet is a general term, I believe in this case "ethereum" is a more correct folder name

src/utils/carbonApi.ts
```diff
@@ -29,26 +30,15 @@ export type RoiRow = {
...
 
 const carbonApi = {
   getCheck: async (): Promise<boolean> => {
-    if (import.meta.env.VITE_DEV_MODE) {
-      return false;
-    }
-    const { data } = await carbonApiAxios.get<boolean>('/check');
+    if (config.mode === 'development') return false;
+    const baseURL = config.appUrl + '/api/';
+    const { data } = await axios.create({ baseURL }).get<boolean>('/check');
```
pingustar, line 41: Just thinking out loud. We could use a new axios instance but with relative path? Then it would work on previews with preview CF functions.

src/pages/privacy/index.tsx
```diff
@@ -583,7 +584,7 @@ export const PrivacyPage = () => {
           THIS PRIVACY POLICY IS SUBJECT IN ALL RESPECTS TO THE TERMS OF USE
           AVAILABLE AT{' '}
           <a
-            href="https://app.carbondefi.xyz/terms"
+            href={config.appUrl + '/terms'}
```
pingustar, line 587: Dont we have these links in the externalLinks object?

src/config/ethereum/common.ts
```diff
@@ -0,0 +1,43 @@
+import { AppConfig } from 'config/types';
+
+export const commonConfig: AppConfig = {
+  mode: 'development',
+  appUrl: 'https://app.carbondefi.xyz',
+  carbonApi: 'https://api.carbondefi.xyz/v1/',
+  rpcUrl:
+    import.meta.env.VITE_CHAIN_RPC_URL ||
+    'https://eth-mainnet.alchemyapi.io/v2/41hlTcJUXX7kLw5oPejnQJ7V9BdW8Ysg',
```
pingustar, line 9: IMHO keys should not be deployed to github. Specially not our dev key for localhost usage. We had abuse in the past of our keys.

src/config/ethereum/common.ts
```diff
@@ -0,0 +1,43 @@
...
+    carbon: {
+      carbonController: '0xC537e898CD774e2dCBa3B14Ea6f34C93d5eA45e1',
+      voucher: '0x3660F04B79751e31128f6378eAC70807e38f554E',
+    },
+    utils: {
+      multicall: '0x5ba1e12693dc8f9c48aad8770482f4739beed696',
+    },
+  },
+  tokenLists: {
+    Bancor: 'https://d1wmp5nysbq9xl.cloudfront.net/ethereum/tokens.json',
```
pingustar, line 40: I would lowercase this

src/config/index.ts
```diff
@@ -0,0 +1,25 @@
+import ethereumDev from './ethereum/development';
+import ethereumProd from './ethereum/production';
+
+const configs = {
```
pingustar, line 4: I would probably move this whole config folder out of `src` folder and put one level up in root

src/config/types.ts
```diff
@@ -0,0 +1,37 @@
+export interface AppConfig {
```
pingustar, line 1: do we need this type? Seems redundant to the actual config. Can we not export as const? and derive the type from it?

```
export const config = {} as const;
export type Config = typeof config;
```

src/pages/privacy/index.tsx
tiagofilipenunes: Should also change the text HTTPS://APP.CARBONDEFI.XYZ/TERMS underneath.

src/pages/terms/index.tsx
```diff
@@ -1072,7 +1073,7 @@ export const TermsPage = () => {
           USES OR FROM PROHIBITED JURISDICTIONS. IF YOU DO NOT AGREE TO ANY OF
           THESE TERMS OF USE OR THE PRIVACY POLICY AVAILABLE AT{' '}
           <a
-            href="https://app.carbondefi.xyz/privacy"
+            href={config.appUrl + '/privacy'}
```
tiagofilipenunes, line 1076: Should also change the text HTTPS://APP.CARBONDEFI.XYZ/PRIVACY underneath

src/config/ethereum/common.ts
```diff
@@ -0,0 +1,43 @@
+import { AppConfig } from 'config/types';
+
+export const commonConfig: AppConfig = {
+  mode: 'development',
+  appUrl: 'https://app.carbondefi.xyz',
+  carbonApi: 'https://api.carbondefi.xyz/v1/',
+  rpcUrl:
+    import.meta.env.VITE_CHAIN_RPC_URL ||
+    'https://eth-mainnet.alchemyapi.io/v2/41hlTcJUXX7kLw5oPejnQJ7V9BdW8Ysg',
+  walletConnectProjectId: 'f9d8863ab6c03f2293d7d56d7c0c0853',
```
tiagofilipenunes, line 10: This id is not currently exposed.

src/config/types.ts
```diff
@@ -0,0 +1,37 @@
...
+      WBTC: string;
+      SHIB: string;
+      ENJ: string;
+      UNI: string;
+      LINK: string;
+      LDO: string;
+      APE: string;
+      GRT: string;
+      AAVE: string;
+      CRV: string;
```
tiagofilipenunes, line 26: If the idea of having a config is to facilitate app interactions with deployments on other chains, this type enforces that all these tokens exist on that chain, which most likely won't be the case

tiagofilipenunes (changes requested):
Great work, will make it much easier to use the app for deployments on other chains!

src/utils/carbonApi.ts
GrandSchtroumpf: I'm not familiar with cloudflare lambda. They are deployed alongside the preview link ?
I think it's much better to use relative path as you mentioned because it's part of the repo.

src/config/index.ts
GrandSchtroumpf: This is what I did at first, but since the absolute path starts at `src` it makes it much easier to use in the app : 
```typescript
import config from 'config';
```

I can try to add config as alias in the tsconfig.json.

src/config/types.ts
GrandSchtroumpf: The main goal of this interface is to make sure all modes have the same keys. I can derive it but I see two issues : 
- One of the config must be the reference (ethereum/production for example).
- `as const` will make string as string literal. So we would have to hard type them in the reference object to guide typescript.
- It makes things a little bit less explicit. Since this config folder is dedicated to external team, I think it's better to explicitly create a readable interface they can look at.

src/config/ethereum/common.ts
GrandSchtroumpf: sure, it was capitalized previously. TBH I'm considering using an array with comments instead of an object since it's not used.

src/config/ethereum/common.ts
GrandSchtroumpf: It is, just not in Github. Since it's in the client, anyone can access it though their console.

src/config/index.ts
pingustar: no need, its fine.

src/services/web3/config.ts
```diff
@@ -1,42 +1,15 @@
...
-  ...conf,
-  carbon: {
-    ...conf.carbon,
+    ...conf.addresses.carbon,
     carbonController:
       lsService.getItem('carbonControllerAddress') ||
-      conf.carbon.carbonController,
-    voucher: lsService.getItem('voucherContractAddress') || conf.carbon.voucher,
+      conf.addresses.carbon.carbonController,
+    voucher:
```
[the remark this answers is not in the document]
GrandSchtroumpf: This can be used by our partners if they use Tenderly while testing their contract. So I'll keep the file for now

src/config/types.ts
GrandSchtroumpf: I created an issue to discuss this: https://github.com/bancorprotocol/carbon-app/issues/1042

src/services/web3/config.ts
```diff
@@ -1,42 +1,15 @@
...
-    ENJ: string;
-    UNI: string;
-    LINK: string;
-    LDO: string;
-    APE: string;
-    GRT: string;
-    AAVE: string;
-    CRV: string;
-  };
+export const config = {
```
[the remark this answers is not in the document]
GrandSchtroumpf: Then I'll keep it out of practicality ^^

pingustar (approved):
Nice work!

tiagofilipenunes (approved):

PR #1047, opened 2024-02-15
Feat: Add Powered by Carbon DeFi footer

tiagofilipenunes:
- Hidden on mobile
- Stays behind modals
- No issues with showing on the bottom of page elements apart from the simulator set-up page


src/App.tsx
```diff
@@ -41,6 +43,7 @@ export const App = () => {
       <ModalProvider />
       <Toaster />
       <CreateStrategyCTAMobile />
+      {!isNativeApp && <Footer />}
```
GrandSchtroumpf, line 46: It's good to put the footer at the root to make sure its landmark is maintain.
I would just have put it just after the main tag, this is the logical HTML flow.

src/components/common/Footer/Footer.tsx
```diff
@@ -0,0 +1,11 @@
+import { FC } from 'react';
+import { ReactComponent as LogoCarbonDeFi } from 'assets/logos/carbondefi.svg';
+
+export const Footer: FC = () => {
+  return (
+    <footer className="bg-body fixed bottom-0 left-0 z-10 hidden h-65 w-full items-center justify-start gap-8 border-t border-darkGrey py-16 px-20 md:flex">
```
GrandSchtroumpf, line 6: I would avoid using fixed positions when possible.
If you don't use fixed, you wont' need the w-full anymore.
I would avoid setting a defined height but use only paddings instead to make it future proof if we want to add more content.
I think the color used for the border is emphasis (#212123 in figma)

src/components/common/Footer/Footer.tsx
```diff
@@ -0,0 +1,11 @@
...
+      <span className="text-14 font-weight-500 text-white/60">Powered by </span>
+      <LogoCarbonDeFi className="h-16 w-[114px]" />
```
GrandSchtroumpf, line 8: I think this should be wrapped in a link.
Since this is a svg, you shoudn't set the width and height, the svg tag will enforce ratio.

src/config/ethereum/common.ts
```diff
@@ -2,6 +2,7 @@ import { AppConfig } from 'config/types';
 
 export const commonConfig: AppConfig = {
   mode: 'development',
+  productionHostName: 'carbondefi.xyz',
```
GrandSchtroumpf, line 5: I think it should be part of the external link and not here, since the goal is to use it when the app is **not** carbon

src/utils/helpers/index.ts
```diff
@@ -4,9 +4,14 @@ import { Pathnames } from 'libs/routing';
...
 import Graphemer from 'graphemer';
 import { TradePair } from 'libs/modals/modals/ModalTradeTokenList';
+import config from 'config';
 export * from './number';
 
 export const isProduction = window
+  ? window.location.host.includes(config.productionHostName)
+  : true;
+
+export const isNativeApp = window
```
GrandSchtroumpf, line 14: I'm not sure what's the best way to handle that, should we see the powerBy on development ?
Maybe having a debug flag might be useful to easily see how it looks like. To be discussed

src/config/ethereum/common.ts
tiagofilipenunes: There is a distinction to be made here though:

isProduction -> used for when the app is in Production, to hide the Debug menu. Therefore, any partner wanting to fork the code should be able to change what their production hostName is.

isNativeApp -> used to check whether the production app is carbondefi.xyz or not. If it is, then show the footer. Therefore, the native app hostname should be hardcoded vs easily changeable by changing the config

GrandSchtroumpf (approved):
looks great. Good work

ashachaf:
- [ ] footer not visible on simulator page (where you create the simulation)
- [ ] since we took a different approach, it should be visible on mobile.

ashachaf:
- [ ] create strategy - after selecting both tokens, we should default select recurring strategy box
https://github.com/bancorprotocol/carbon-app/assets/40430285/4578f333-3938-4c6d-ba4c-1b34704966c3

- [ ] footer layout on mobile is off
iphone 13mini
image

ashachaf (approved):

PR #1070, opened 2024-02-21
Add simulator tooltips

tiagofilipenunes:
Simulator input page
- Add token selection tooltip
- Add strategy type tooltips

Simulator result page
- Remove TBD from ROI tooltip
- Add Estimated Gains tooltip

### DIFF
--- a/src/components/simulator/input/SimInputTokenSelection.tsx
+++ b/src/components/simulator/input/SimInputTokenSelection.tsx
@@ -39,7 +39,17 @@ export const SimInputTokenSelection: FC<Props> = ({
         <h2 className="m-0 text-18 font-weight-500">Token Pair</h2>
         <Tooltip
           iconClassName="h-18 w-18 text-white/60"
-          element={<div>TBD</div>}
+          element={
+            <div>
+              Selecting the tokens you would like to create a simulation for.
+              <br />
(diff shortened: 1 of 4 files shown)


pingustar:
keep it up

src/components/simulator/input/SimInputTokenSelection.tsx
```diff
@@ -39,7 +39,17 @@ export const SimInputTokenSelection: FC<Props> = ({
         <h2 className="m-0 text-18 font-weight-500">Token Pair</h2>
         <Tooltip
           iconClassName="h-18 w-18 text-white/60"
-          element={<div>TBD</div>}
+          element={
+            <div>
```
GrandSchtroumpf, line 43: Use `p` instead of `div` when you've got only text

src/components/simulator/result/SimResultSummaryGains.tsx
```diff
@@ -49,6 +49,9 @@ export const SimResultSummaryGains: FC<Props> = ({
 
 const TooltipContent: FC<{}> = () => (
   <>
-    <span className="align-middle">TBD.&nbsp;</span>
+    <span className="align-middle">
```
GrandSchtroumpf, line 52: I think `p` would make more sense here since this is a block.

GrandSchtroumpf (approved):

PR #1084, opened 2024-02-28
[Simulator] Bugs fixes

### LINKED ISSUE #1078: If there is no price chart, user can still click on “Start Simulation” button
open simulator page
    select two tokens that have no price chart history(00 and 0xagon for example)
    expected: The following design is implemented and a user can't start the simulation [mockup](https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=13105%3A108243&mode=design&t=w0jvvTf11J063fsw-1)
    actual: The above design isn't implemented and user can click on “Start Simulation” CTA

GrandSchtroumpf:
fix #1078
- [x] Send the UTC version of the day to the server
- [x] Transform back to local timezone from server
- [x] Implement UI when the token pair has no history price

### DIFF
--- a/src/components/common/datePicker/DateRangePicker.tsx
+++ b/src/components/common/datePicker/DateRangePicker.tsx
@@ -2,10 +2,10 @@ import { Button } from 'components/common/button';
 import { Calendar, CalendarProps } from 'components/common/calendar';
 import { DropdownMenu, MenuButtonProps } from 'components/common/dropdownMenu';
 import { subDays, isSameDay, subMonths } from 'date-fns';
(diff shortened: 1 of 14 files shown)


ashachaf:
- [ ] date picker - pre select not visible
flow:
1. run simulation
2. click the back option
3. open date picker
expected: if the selected range falls into one of the pre defined option, it should be highlighted
actual: it is not highlighted

- [ ] missing price history text
replace "token" with "pair"
image

zavelevsky (approved):
code LGTM

PR #1105, opened 2024-03-05
Avoid scroll on tooltip

### LINKED ISSUE #1104: Download tooltip on simulation result page
### How to reproduce
On simulation result page mouse hover over the download icon

- [ ] Tooltip opacity
expected: The tooltip should be non-transparent.
actual: The color of the "Create strategy" button is visible through the tooltip

- [ ] Horizontal scrolling
expected: Display a tooltip when hovering over the download icon without altering the screen width.
actual: Hovering over the download icon causes a temporary horizontal scroll on the page, disappearing after a second.

GrandSchtroumpf:
fix #1104 
Remove bounce effect for the download tooltip animation


pingustar (approved):
Code LGTM

Issue #1040, opened 2024-02-13
Use native fetch instead of Axios


GrandSchtroumpf:
We currently use the axios library to call the server. Axios is not needed anymore since `fetch` is natively supported by all browsers and node versions.

PR #1110, opened 2024-03-06
Issue #1040 - Replace axios with fetch

tiagofilipenunes:
- [x] Remove Cryptocompare API module as it is not used in the app
- [x] Replace axios with fetch in tokenHelper and carbonApi
- [x] Remove axios dependency

https://caniuse.com/fetch


src/libs/tokens/tokenHelperFn.ts
```diff
@@ -15,10 +14,14 @@ const buildIpfsUri = (ipfsHash: string) => `https://ipfs.io/ipfs/${ipfsHash}`;
 export const fetchTokenLists = async () => {
   const res = await Promise.all(
     config.tokenLists.map(async (uri) => {
-      const res = await axios.get<TokenList>(uri, { timeout: 10000 });
+      const signal = AbortSignal.timeout(10000);
+      const res: TokenList = await fetch(uri, {
+        method: 'GET',
+        signal,
+      }).then((response) => response.json());
```
GrandSchtroumpf, line 21: You don't need the method: "GET", this is the default.
Also you need to check for error : 
```suggestion
      const response = await fetch(uri, { signal });
      if (!response.ok) throw new Error(....);
      const result: TokenList  = await response.json());
```

src/utils/carbonApi.ts
```diff
@@ -32,54 +31,55 @@ export type RoiRow = {
...
+  }
+  const response = await fetch(url);
+  const result = (await response.json()) as T & { error?: string };
+
+  if (!response.ok)
+    throw new Error(
+      result?.error ||
+        `Response was not okay. ${response.statusText} response received.`
+    );
+  return result;
```
GrandSchtroumpf, line 47:
```suggestion
  const result = await response.json()

  if (!response.ok) {
    const error = (result as { error?: string }).error;
    throw new Error(
      error ||
        `Response was not okay. ${response.statusText} response received.`
    );
  }
  return result as T;
```

src/utils/carbonApi.ts
```diff
@@ -32,54 +31,55 @@ export type RoiRow = {
...
     return data;
   },
   getMarketRateHistory: async (
     params: TokenPriceHistorySearch
   ): Promise<TokenPriceHistoryResult[]> => {
-    const { data } = await newApiAxios.get<TokenPriceHistoryResult[]>(
-      `history/prices`,
-      { params }
-    );
+    const data = await get<TokenPriceHistoryResult[]>('history/prices', params);
```
GrandSchtroumpf, line 68:
```suggestion
    return get<TokenPriceHistoryResult[]>('history/prices', params);
```

src/utils/carbonApi.ts
```diff
@@ -32,54 +31,55 @@ export type RoiRow = {
...
     return data;
   },
   getRoi: async (): Promise<RoiRow[]> => {
-    const { data } = await newApiAxios.get<RoiRow[]>('roi');
+    const data = await get<RoiRow[]>('roi');
```
GrandSchtroumpf, line 72:
```suggestion
    return get<RoiRow[]>('roi');
```

GrandSchtroumpf (approved):
Nice work

Issue #1098, opened 2024-03-04
[Explorer] Wrong token link on mobile


GrandSchtroumpf:
On mobile the table with the tokens redirect to /strategies/... while they should redirect to /explorer/...
![image](https://github.com/bancorprotocol/carbon-app/assets/8143464/41f09159-a7ab-4594-8bd9-7b0342cb659c)

PR #1115, opened 2024-03-08
Issue #1098 - Fix portfolio mobile table redirecting to wrong link

tiagofilipenunes:
- Move GetPortfolioTokenHref to separate types file
- Fix explorer and strategies portfolio pages to direct to the correct portfolio token link.


src/pages/strategies/portfolio/index.tsx
```diff
@@ -1,26 +1,23 @@
...
+
+  const getPortfolioTokenHref: GetPortfolioTokenHref = (row) => ({
+    href,
+    params: { address: row.token.address },
+  });
 
   const onRowClick = (row: Row<PortfolioData>) =>
     navigate({
-      to: '/strategies/portfolio/token/$address',
+      to: href,
```
GrandSchtroumpf, line 20: I think you're loosing some typing here. Check with @pingustar what the best strategy here with the router.

tiagofilipenunes: Changing `href` throws a type error

image

GrandSchtroumpf (approved):

Issue #957, opened 2023-12-28
Strategy deletion popup


Costa769:
# Designs
[Popup to suggest strategy deletion](https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=11922-51834&mode=design&t=VeoeBD7rCC9c4G2L-4)

# Overview

When a user withdraws the entire budget allocated to a strategy, we want to provide them with the option to delete that strategy. This way, we can avoid maintaining a list of inactive strategies and keep the platform more organized.

**Reminder**
The Withdrawal page is accessed by clicking on the "Withdraw funds" option in the overview card manage menu.

## Flow

1. User opens a withdraw funds page
2. User enters the amount to withdraw
3. User clicks on the “Withdraw” button
4. On “Withdraw” CTA click check the following:
- **IF** both budgets are fully withdrawn( both budgets=0)
-- Show the popup to suggest strategy deletion
- **ELSE, IF** at least one of the budgets >0
-- No change to the current flow(proceed with the withdrawal)

### Popup actions:

- On "X" icon click 
-- Close the popup
- On click outside the popup 
-- Close the popup.
- On “Withdraw only” click
-- Proceed with funds withdrawal flow (once the transaction is confirmed in the wallet, redirect the user to My strategies overview page)
- On "Withdraw and Delete" click:
-- Trigger the "deleteStrategy" function (once the transaction is confirmed in the wallet, redirect the user to My strategies overview page)

### Empty strategy warning
Remove the following warning from the overlapping strategy withdraw funds page. This warning is displayed when user tries to withdraw the entire strategie's budget
"Please note that your strategy will be inactive as it will not have any budget."

Issue #1124, opened 2024-03-13
Add Simulate strategy option to strategy menu


Costa769:
### Overview
We aim to provide users with the capability to initiate a strategy simulation directly from the strategy card management menu.

**Note**
Users can enter the strategy card management menu by navigating to either the My Strategies or Explorer pages and clicking on the gear icon  on the strategy card.

### Logic

IF the strategy is `recurring` or `overlapping`, add "Simulate Strategy" option into the strategy card management menu on both the My Strategies and Explorer pages.

- The "Simulate Strategy" option should exclusively appear only in the menu for recurring and overlapping strategies.
- Place this option below the "Manage Notifications" option.
- Simulate strategy tooltip "Simulate the strategy's performance based on historical data"
- Upon clicking "Simulate Strategy"
-- Open the simulate page (in the same tab) with all the strategy settings prefilled to match the chosen strategy current data for simulation.
-- Ensure that all relevant errors and warning messages are displayed as applicable. 

<img width="518" alt="image" src="https://github.com/bancorprotocol/carbon-app/assets/40430285/73f02d92-e703-4180-9330-d92cd161d5d9">

PR #1137, opened 2024-03-25
Add simulator overlapping

pingustar:
This PR fixes:

#957 PR: #1164 
#1124 PR: #1160 

and overlapping simulator feature


ashachaf:
- [ ] first visit - missing spread
on first visit, the spread shows NaN and the chart is missing the selected area
image

- [ ] budget inputs
- when the selected range is above/below the price of the first point, one of the budgets should be locked from user inputs.
- when the price of the first point is within the selected rage:
-- input budget -> calculate the 2nd budget required
-- change prices on the chart/input 
-> expected: update the budget calculation using the last user input.
-> actual: budget does not update -> this will lead into incorrect budgets and distribution on the range (invalid)

- [ ] (optional change) spread
- select 0.1%
- input 0 in the custom field
expected: spread goes back to the last selected option
actual: spread goes back to default value (in this case 0.05%)

- [ ] chart drag doesnt work
expected: click and drag the selected range
actual: click and drag doesnt work

- [ ] simulation error "Missing Information"
https://197d60d9.carbon-app-csq.pages.dev/simulate/overlapping?baseToken=0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee&quoteToken=0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48&buyMin=3333.362388&buyMax=3744.556883349950149551345962113659022931206380857427716849451645064805583250249252243270189431704885&buyBudget=&buyIsRange=true&buyMarginal=3333.362388&sellMin=3343.362475164&sellMax=3755.790554&sellBudget=&sellIsRange=true&sellMarginal=3343.362475164&start=1708870238&end=1711375838&overlappingSpread=0.3&type=overlapping

- [ ]

ashachaf:
- [ ] update budget information message text
- Sell budget message:
The first price in the selected time frame is outside the ranges you've set for selling {Base}. Therefore, budget for buying {Base} is not required. Learn More
- Buy budget message:
The first price in the selected time frame is outside the ranges you've set for buying {Base}. Therefore, budget for selling {Base} is not required. Learn More

- [ ] price diff from market price indication
currently, there is an indication of diff from market price below the min and max price inputs.
please remove this indication on the simulation page.

- [ ] sell budget always on top

- [ ] create strategy cta
currently: opens the first page of the create strategy flow
updated: expected to open the 2nd page and pre-fill the price and spread fields according to user inputs

ashachaf:
- [ ] budget input field title
in the budget section, the two input fields are missing titles to clearly indicate which input is which.
please add titles to each budget input field.
example:
image
follow the design already in place found [here](https://app.carbondefi.xyz/strategies/create?base=0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE&quote=0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48&strategyType=recurring&strategySettings=range)

- [ ] create overlapping page - update budget input fields
on the create page for overlapping strategy, update the following:
- add title to the budget input fields
- position the sell as the top option

src/libs/modals/modals/ModalWithdrawOrDelete.tsx
```diff
@@ -0,0 +1,53 @@
...
+    <ModalOrMobileSheet id={id}>
+      <div className="my-20">
+        <IconTitleText
+          variant={'error'}
+          icon={<IconDelete />}
+          title={
+            'This strategy will become inactive once the budget is removed'
+          }
+          text={'Delete this strategy to keep things tidy'}
+        />
```
GrandSchtroumpf, line 29:
```suggestion
        <IconTitleText
          variant="error"
          icon={<IconDelete />}
          title="This strategy will become inactive once the budget is removed"
          text="Delete this strategy to keep things tidy"
        />
```

src/libs/queries/extApi/simulator.ts
```diff
@@ -56,45 +56,70 @@ export type SimulatorReturn = {
...
+export const useGetSimulator = (search: SimulatorResultSearch) => {
   return useQuery<SimulatorReturn, Error>(
-    QueryKey.simulator(params),
+    QueryKey.simulator(search),
     async () => {
       try {
+        const { buyIsRange, sellIsRange, type, spread, ...params } = search;
         const res = await carbonApi.getSimulator(params);
 
+        // TODO cleanup schema
```
GrandSchtroumpf, line 100: If you don't do it in this PR, can you open an issue with reference the issue id in the TODO comment ?

src/libs/routing/router.tsx
```diff
@@ -1,18 +1,43 @@
...
-    return Object.fromEntries(searchParams.entries());
-  },
-  stringifySearch: (search) => {
-    const searchParams = new URLSearchParams();
-    for (const key in search) searchParams.set(key, search[key]);
-    const searchStr = searchParams.toString();
-    return searchStr ? `?${searchStr}` : '';
-  },
+  parseSearch: parseSearchWith(JSON.parse),
+  // parseSearch: (searchStr) => {
```
GrandSchtroumpf, line 8: Can we remove ?

src/libs/routing/routes/sim.tsx
```diff
@@ -163,11 +212,17 @@ export const simulatorResultRoute = new Route({
       sellMax: search.sellMax,
       sellMin: search.sellMin,
       sellBudget: search.sellBudget || '0',
-      sellIsRange: stringToBoolean(search.sellIsRange),
+      // TODO add validation
```
GrandSchtroumpf, line 215: Same as above

src/components/strategies/common/BudgetInput.tsx
```diff
@@ -8,44 +8,62 @@ import { UseQueryResult } from '@tanstack/react-query';
...
+    title,
+    titleTooltip,
+  } = props;
   const inputId = useId();
-  const balance = query.data ?? '0';
+  const balance = query?.data ?? '0';
 
   return (
     <div className="flex flex-col gap-16">
+      <legend className="flex text-14 font-weight-500">
```
GrandSchtroumpf, line 40: legend must be used inside a <fieldset>
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/legend

Maybe use a label instead ?

src/components/strategies/common/BudgetInput.tsx
```diff
@@ -8,44 +8,62 @@ import { UseQueryResult } from '@tanstack/react-query';
 interface Props {
   id?: string;
   children?: ReactNode;
-  order: OrderCreate;
+  budgetValue: string;
+  budgetError?: string;
```
GrandSchtroumpf, line 12: Since the name of the component is `BudgetInput`, I think we can rename to `value` and `error`

src/components/simulator/input/SimInputStrategyType.tsx
```diff
@@ -46,33 +47,38 @@ export const SimInputStrategyType: FC<Props> = ({ strategyType }) => {
             role="tab"
             id={'tab-' + label}
             aria-controls={'panel-' + label}
-            aria-selected={strategyType === label}
             key={label}
-            to="/simulate/$simulationType"
+            to={`/simulate/${label}`}
```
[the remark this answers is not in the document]
pingustar: I refactored this to be two separate routes, so there is no `$simulationType` param anymore. But type safety is still working incase `label` is changed to something other than "recurring" or "overlapping"

src/components/simulator/input/overlapping/CreateOverlappingStrategy.tsx
```diff
@@ -0,0 +1,295 @@
...
+        const { min, max } = getInitialPrices(marketPrice);
+        setOverlappingParams(min, max);
+      });
+    } else {
+      setOverlappingParams(state.buy.min, state.sell.max);
+    }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
+  }, [marketPrice, spread]);
+
+  const setErrorCb = useCallback(
```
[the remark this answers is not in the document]
pingustar: without this being a callback, the useEffect in the child where we pass this down to, is triggered on every render causing a loop.

src/components/strategies/create/overlapping/CreateOverlappingStrategyBudget.tsx
```diff
@@ -39,9 +40,14 @@ export const CreateOverlappingStrategyBudget: FC<Props> = (props) => {
...
   const buyBudgetId = useId();
   const sellBudgetId = useId();
+  const { user } = useWeb3();
 
-  const checkInsufficientBalance = (balance: string, order: OrderCreate) => {
-    if (new SafeDecimal(balance).lt(order.budget)) {
+  const checkInsufficientBalance = (
+    balance: string,
+    order: OrderCreate,
+    user?: string
```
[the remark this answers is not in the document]
pingustar: the useEffect needs to be triggered on user change, to make it clear that this a dependency required I added it as a param. These are not callbacks and the useEffect has eslint hints disabled making it very prone for errors otherwise

src/components/strategies/edit/overlapping/WithdrawOverlappingStrategy.tsx
```diff
@@ -210,39 +214,23 @@ export const WithdrawOverlappingStrategy: FC<Props> = (props) => {
...
-              </span>
-            </output>
-          )}
-        </footer>
+        {!withdrawAll && (
+          <footer className="flex items-center gap-8">
+            <IconAction className="h-16 w-16" />
+            <p className="text-12 text-white/60">
+              Price range and liquidity spread remain unchanged.&nbsp;
+              <a
```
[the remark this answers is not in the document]
pingustar: dont recall touching this, but will fix

src/components/simulator/input/overlapping/CreateOverlappingStrategy.tsx
```diff
@@ -0,0 +1,295 @@
...
+    base,
+    quote,
+    order: {
+      min: state.buy.min,
+      max: state.sell.max,
+      price: '',
+      isRange: true,
+    },
+    buy: true,
+  });
```
tiagofilipenunes, line 58: I think this isn't being used down the tree here, might be worth removing

tiagofilipenunes (changes requested):
Observed while reviewing (feel free to ignore):
- Because the result page is not keeping track of overlapping buy and sell budgets:
  - When simulating an overlapping strategy and going back, the buy and sell budgets are cleared in the simulator input page
  - Creating a strategy from the overlapping simulator doesn't populate budgets
- Swapping tokens in simulator input clear the buy/sell budget
- Cannot change simulation dates in overlapping results page, but can change them in the input. While I understand this is a deliberate choice, it is possible to set dates in the simulation input which makes the behaviour a bit inconsistent. An issue to enable the calendar for simulator result page in a separate PR could be created

Bug:
- After swapping to overlapping from recurring, then changing token, the prices shows N/A
  1. Go to simulator input recurring, pick something like ETH/USDT
  2. Change to overlapping, the token choice stayed
  3. Change the quote token and the price chart shows N/A

tiagofilipenunes (approved):

GrandSchtroumpf (approved):
Great work !

PR #1144, opened 2024-03-28
Create input to change backend url

### LINKED ISSUE #1135: [Debug] Add backend url input
**Set URL** :
- [ ] Add an input in debug page to change backend url
- [ ] Keep URL in localstorage
- [ ] Update carbonAPI to get the URL from localstorage

**Get URL** :
Either one or the other :
1. Refresh page when backend url changes
2. Listen on [storage event](https://developer.mozilla.org/en-US/docs/Web/API/StorageEvent) from localstorage to update carbonAPI

**Bonus**: 
If backend url can be derived from the forkId, fill the backend url when we set rpc url

GrandSchtroumpf:
fix #1135 

- [x] add backend url in localstorage
- [x] use localstorage in config


tiagofilipenunes (changes requested):
Looks good, just one comment 👍

- [ ] Would if make sense to move the dependency on the localStorage by the src/config/ethereum/common.ts away from this config file to make it as simple as possible to change for forks?

tiagofilipenunes (approved):

PR #1146, opened 2024-03-28
Export full id from activity

### LINKED ISSUE #1145: Make Full ID available in activity

GrandSchtroumpf:
fix #1145

### DIFF
--- a/src/components/activity/ActivityExport.tsx
+++ b/src/components/activity/ActivityExport.tsx
@@ -6,7 +6,8 @@ import { ReactComponent as IconDownloadFile } from 'assets/icons/download-file.s
 
 const getCSV = (activities: Activity[]) => {
   const header = [
-    'ID',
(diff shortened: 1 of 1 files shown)


tiagofilipenunes (changes requested):
Great UX improvements!

- [x] Selecting "Deposit" and "Delete" actions in the Activity Table shift the columns. Is this behaviour expected?


https://github.com/bancorprotocol/carbon-app/assets/66181828/29b5d9fb-0bda-4bcb-9257-e51f9ab8be34



- @Costa769  Since now a user can easily select a strategy that has already been deleted from the activity table or list, I suggest a change to the message for strategy not found to change from 

'The strategy you are looking for does not exist.'
to
'The strategy you are looking for does not exist or has been deleted.'

tiagofilipenunes (approved):

PR #1157, opened 2024-04-04
[Activity] Notifications

### LINKED ISSUE #1147: Activity notifications
### Overview
We aim to integrate trade notifications into the app. This entails displaying a notification to users when a trade occurs against one of their strategies while they're connected to Carbon application

- Notifications will be displayed on all app.carbondefi.xyz pages upon user wallet connection. 
- These notifications will exclusively pertain to "buy low" and "sell high" strategy activities.
- To implement this feature, utilize the existing Activity API to fetch data (https://api.carbondefi.xyz/v1/activity). No additional data beyond what is provided by the existing activity API is required.

 **Logic**
 - **IF** owner = connected wallet and  action = "buy" or "sell" 
-- Display the relevant notification for 7 seconds before hiding it [mockup](https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=11573%3A40361&mode=design&t=CVCMCu654JGpBekG-1)
-- Clicking on the "x" icon will dismiss the notification.
-- Clicking on the notification itself will redirect the user to the corresponding strategy activity page.

GrandSchtroumpf:
fix #1147 

How to test : 
1. Create a strategy
2. Trade against it
3. under 1min you should see a notification

### DIFF
--- a/src/components/activity/ActivityNotification.tsx
+++ b/src/components/activity/ActivityNotification.tsx
@@ -1,0 +1,66 @@
+import { FC, useId } from 'react';
+import { useNotifications } from 'hooks/useNotifications';
+import { NotificationActivity } from 'libs/notifications/types';
+import { activityActionName, activityDescription } from './utils';
+import { ActivityIcon, ActivityId } from './ActivityTable';
+import { ReactComponent as IconClose } from 'assets/icons/X.svg';
+import { useInterval } from 'hooks/useInterval';
+import { FOUR_SECONDS_IN_MS } from 'utils/time';
+import { Link } from '@tanstack/react-router';
+
+interface Props {
+  notification: NotificationActivity;
+  isAlert?: boolean;
+}
+
+export const ActivityNotification: FC<Props> = ({ notification, isAlert }) => {
+  const titleId = useId();
+  const { activity } = notification;
+  const { dismissAlert, removeNotification } = useNotifications();
+
+  const handleCloseClick = () => {
+    if (isAlert) {
+      dismissAlert(notification.id);
+    } else {
+      removeNotification(notification.id);
+    }
+  };
+
+  useInterval(
+    () => dismissAlert(notification.id),
+    isAlert ? FOUR_SECONDS_IN_MS : null,
+    false
+  );
+
+  return (
+    <article aria-labelledby={titleId} className="flex gap-16">
+      <ActivityId activity={activity} size={16} />
+      <Link
+        to="/strategy/$id"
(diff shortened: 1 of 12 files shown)


src/components/activity/useActivityNotifiction.ts
```diff
@@ -0,0 +1,27 @@
...
+  const [previous, setPrevious] = useState<number | null>(null);
+  const query = useActivityQuery({ ownerId: user });
+  const allActivities = query.data || [];
+  const buyOrSell = allActivities.filter(
+    (a) => a.action === 'sell' || a.action === 'buy'
+  );
+  const { dispatchNotification } = useNotifications();
+
+  useEffect(() => {
+    if (query.isLoading) return;
```
pingustar, line 17: this looks like something that the query callbacks would handle great - wanna jump on a quick session to explore if we can refactor this to remove the useEffect and use the useQuery callbacks?

src/components/activity/useActivityQuery.tsx
```diff
@@ -50,7 +57,7 @@ export const useActivityQuery = (params: QueryActivityParams = {}) => {
       });
     },
     {
-      enabled: !isLoading,
+      enabled: !isLoading && validParams,
       refetchInterval: 30 * 1000,
```
pingustar, line 61: nit: I think we have a 30sec const

src/components/activity/ActivityNotification.tsx
```diff
@@ -0,0 +1,66 @@
...
+import { useInterval } from 'hooks/useInterval';
+import { FOUR_SECONDS_IN_MS } from 'utils/time';
+import { Link } from '@tanstack/react-router';
+
+interface Props {
+  notification: NotificationActivity;
+  isAlert?: boolean;
+}
+
+export const ActivityNotification: FC<Props> = ({ notification, isAlert }) => {
```
pingustar, line 16: nit: This component looks like a duplicate of the existing component in terms of logic - maybe we can extract this into a base component and then just pass the markup down. If the base component returns a the children as a function, then you can access the `handleCloseClick` in the parents. Not sure if I explained this correct - shoot me a msg if not clear what I am trying to say :D

pingustar (approved):
LGTM

src/components/activity/ActivityNotification.tsx
GrandSchtroumpf: The feature is deprecated in v5.

PR #1166, opened 2024-04-13
[Hotfix] Migrate notification from LS

GrandSchtroumpf:
Force type to "tx" if notification in LS doesn't have a type


pingustar (approved):
LGTM

PR #1177, opened 2024-04-18
Improve E2E stability

### LINKED ISSUE #967: [E2E] Optimizations
- [ ] Use yarn cache in Github Action
- [ ] Use Docker image with playwright browser in Github Action (https://github.com/microsoft/playwright/issues/23388#issuecomment-1570016619)

### LINKED ISSUE #1004: Improve e2e screenshot stability
Two issues in this [PR](https://github.com/bancorprotocol/carbon-app/pull/1003) with screenshots:

1. Some screenshots are showing visual diff when there are small pixel changes
2. When some screenshots are taken, the cursor may be on top of a tooltip, showing it as part of the screenshot

Goal is to make sure that screenshots are only committed when there are visual diffs (as much as possible)

### LINKED ISSUE #1093: [Simulator] Set an initial cut-off date for the picker
Set an initial cut-off date for the picker. The date picker allows to select until one year from today, but not before that. This means that previously available date ranges will no longer be available once the start date is more than 1 year old, which also means that the e2e tests will fail after sufficient time as passed as the testcase date ranges can no longer be chosen.

The simulator API can take start dates earlier than 1 year ago while the history prices cannot.

### LINKED ISSUE #1117: [E2E] Remove isE2E flag
Playwright v1.41 allows to set custom styles when taking screenshots, so the isE2E flag can be removed - to reduce the influence of the E2E tests on the implementation.

- [ ] Change datepicker testing implementation so that datePicker doesn't need to use the e2e debug flag
  - By changing the mocked data and the simulator testcases
    - Check current date and receive from testcase the duration from today that we want to test (e.g. `duration="30 days"`)
    - Select the date ranges from Date.now() - duration to Date.now()
    - Mocked data creates the pricing data and simulator data using the correct timestamps
  - By using [style](https://playwright.dev/docs/api/class-page#page-screenshot-option-style) 
- [ ] Update playwright to at least version 1.41 to support screenshots with custom styles
- [ ] Use [style](https://playwright.dev/docs/api/class-page#page-screenshot-option-style) when taking a screenshot so that it is not necessary to set the font in the main wallet button to mono
- [ ] Remove isE2E flag

tiagofilipenunes:
## Related issues

Fix #1117
Fix #1093 
Fix #1004
Fix #967

## Done

- [x] Improve readme with more e2e tips
- [x] Change retention day of upload-artifacts job to 3 days to match repo settings (repo settings must be changed to allow increase)

#1117 & #1093
- [x] Update playwright
- [x] Style user wallet on screenshot
- [x] Remove user wallet isE2E flag
- [x] Mock date and timezone to fix calendar issue with 1-year window and to guarantee charts always look the same
- [x] Remove isE2E flag from calendar
- [x] Remove isE2E logic from app

#967
- [x] Update Github action to use Playwright Docker
- [ ] Yarn cache

#1004
- [x] Fix mock api for new simulator endpoint and mock api script to get data
- [x] Force 1 worker in CI as per playwright recommendations
- [x] Disable animations when taking screenshot
- [x] Disable button hover before screenshots
- [x] Using png instead of jpg for lossless compression
- [x] Mock sentry and GTM routes to abort as not needed for e2e test
- [x] Explore e2e tests stability improvements related to fails dependent on Tenderly
  - Timing out when waiting for tenderly RPC response
  - [x] Implement waitForTenderlyRPC after every form submit
  - [x] Explore other solutions to mock mainnet - implementation is a bigger undertaking and out of scope for this PR
    - Tenderly devnet would save time setting up with the debug page, but won't improve stability
    - [DePayFi web3 mock](https://github.com/DePayFi/web3-mock) doesn't support all chains that CarbonDeFi and forks are but would improve stability
    - [ETH-testing-monorepo](https://github.com/VGLoic/eth-testing-monorepo) is another solution that only works for Ethereum, used by Safe, GMX and CELO. Would also improve stability
- [x] Mock token lists


e2e/utils/operators.ts
```diff
@@ -8,11 +8,18 @@ export const shouldTakeScreenshot = isCI && !isDraft;
 /** Utils to take screenshot on CI that are not draft */
 export const screenshot = (target: Page | Locator, name: string) => {
   if (!shouldTakeScreenshot) return;
+
+  const styles = `
```
GrandSchtroumpf, line 12: You can do that : 
```js
this.page.getByTestId('user-wallet').evaluate((el) => {
  el.style.fontFamily: 'monospace';
  el.style.visibility = 'hidden';
});
```

scripts/mock-api.mjs
```diff
@@ -278,64 +285,77 @@ const getSimulatorData = async (simulatorData) => {
 };
 
 async function main() {
-  const roi = await getRoi();
+  // const roi = await getRoi();
```
GrandSchtroumpf, line 288: Can we remove the comment ?

GrandSchtroumpf (approved):

Issue #1173, opened 2024-04-17
Analytics Preparation


pingustar:
As a prerequisite for the analytics features, we need to update the data fetching and SDK initialisation process to only fetch data / init SDK if on a relevant page.

PR #1184, opened 2024-04-22
#1173 - Analytics Preparation

pingustar:
#1173 

dont init SDK when on defined route. Example PR with `/simulate` route


src/App.tsx
```diff
@@ -9,18 +10,32 @@ import { useStore } from 'store';
 import { Toaster } from 'components/common/Toaster/Toaster';
 import { Footer } from 'components/common/Footer/Footer';
 
-let didInit = false;
+let didInitCheck = false;
```
GrandSchtroumpf, line 13: Out of curiosity, is there a reason to put it here over in a state ?

GrandSchtroumpf (approved):

src/App.tsx
pingustar: with react strict mode disabled, the effect would run twice when the variable is in state.

PR #1197, opened 2024-05-03
Specify parameters in useGetTokenPriceHistory

### LINKED ISSUE #1196: [Simulator] Simulator is requesting token-price-history twice
Actual:

Simulator is requesting token-price-history twice with different parameters

Expected:

One query for the token-price-history with same parameters

image

tiagofilipenunes:
Fix #1196


src/pages/simulator/index.tsx
```diff
@@ -14,7 +14,12 @@ export const defaultEnd = () => getUnixTime(new Date());
 export const SimulatorPage = () => {
   useSimDisclaimer();
   const searchState = simulatorInputRootRoute.useSearch();
-  const { isError } = useGetTokenPriceHistory(searchState);
+  const { isError } = useGetTokenPriceHistory({
```
GrandSchtroumpf, line 17: Couldn't we pass the whole object ?

GrandSchtroumpf (approved):

PR #1214, opened 2024-05-10
fix sentry dsn

pingustar:


src/config/ethereum/production.ts
```diff
@@ -4,7 +4,7 @@ import { commonConfig } from './common';
 const config: AppConfig = {
   ...commonConfig,
   mode: 'production',
-  sentryDNS:
```
GrandSchtroumpf, line 7: You'll need to update sei config & init-sentry to use the new naming

PR #1233, opened 2024-05-24
[Overlapping] Force checkbox on inactive edit price

### LINKED ISSUE #1226: Edit overlapping strategy - if nothing has changed, keep the CTA disabled
flow:
- pick an overlapping strategy
- click the edit prices/deposit/withdraw options
- select anchor token and mark the checkbox
actual: CTA is now enable allowing to submit the change transaction
expected: IF nothing has changed (in any of the market price, buy/sell range or spread), the CTA should remain disabled.

GrandSchtroumpf:
fix #1226 
When editing price of inactive strategy it force recalculation of marginal price, so the CTA is enable even without user change.
By displaying a warning we force the user to check the checkbox before validating CTA


pingustar (approved):
Code LGTM

ashachaf (approved):

PR #1234, opened 2024-05-27
Deploy Carbon app to SEI EVM mainnet

tiagofilipenunes:
- Change config to add Sei Mainnet RPC url, token addresses, popular tokens, default trading pair, network name
- Support adding headers to rpc calls for the network connector and the Carbon SDK


pingustar (approved):
LGTM

zavelevsky (approved):

PR #1236, opened 2024-05-28
[Overlapping] Support "." as input

### LINKED ISSUE #1218: Simulator - handling input values in overlapping strategies
current: 
under "overlapping" only, if you input a value that starts with a decimal point (i.e. .1 instead of 0.1), there is an error

update:
apply the same input number validation as we have in recurring flow.

https://github.com/bancorprotocol/carbon-app/assets/40430285/1ee331ce-2b76-4c2b-aa99-72ed8236eafe

GrandSchtroumpf:
fix #1218 
Support "." as in input for overlapping


tiagofilipenunes (approved):
Great work, LGTM, couldn't find any issues with '.' in input fields, even the debug create strategy is covered 👍

Tried:
- Simulator overlapping range/spread/budget and recurring range/limit sell/buy
- Overlapping strategy create range/spread/budget
- Overlapping strategy edit range/spread/deposit/withdraw
- Overlapping strategy deposit and withdraw budgets

ashachaf (approved):

PR #1271, opened 2024-06-17
use checkbox input for switch

### LINKED ISSUE #1270: [DEV] Use input checkbox for switch
Currently the <Switch /> component is using a <div>, so it cannot be focused nor labelled.
It would be better to use an <input /> tag for that.

GrandSchtroumpf:
fix #1270 
Use input for the <Switch /> component : 
- Support focus
- Check when clicking on label


src/components/common/switch/index.tsx
```diff
@@ -21,22 +21,36 @@ export const Switch: FC<SwitchProps> = ({
...
+        <motion.div
+          className={cn(
+            'aspect-square h-10 rounded-full outline-2',
+            isOn ? 'bg-black outline-black' : 'bg-white/60 outline-white/60',
+            'peer-focus-visible:outline'
+          )}
+          layout
+          transition={spring}
+        />
+      </div>
```
tiagofilipenunes, line 52:
```suggestion
      <div
        className={cn(
          'relative',
          switchStyles({ variant, size, isOn, class: className }),
          'p-0'
        )}
        {...props}
      >
        <input
          className="peer absolute !w-full cursor-pointer opacity-0"
          type="checkbox"
          checked={isOn}
          id={id}
          onChange={(e) => setIsOn(e.target.checked)}
        />
        <motion.div
          className={cn(
            'm-3 aspect-square h-10 rounded-full outline-2',
            isOn ? 'bg-black outline-black' : 'bg-white/60 outline-white/60',
            'peer-focus-visible:outline'
          )}
          layout
          transition={spring}
        />
      </div>
```
- w-full is not being applied as it's being overridden by the checkbox style defined in index.css
- the checkbox is not occupying the size of the switch here so there's an area that is not clickable

tiagofilipenunes (changes requested):
LGTM - just a comment on the size of the checkbox with opacity-0 not matching the size and position of the parent div which led to part of the switch area being unclickable

tiagofilipenunes (approved):

PR #1275, opened 2024-06-18
[Overlapping] Use Price input instead of budget input for market price

### LINKED ISSUE #1224: Edit market price - support price with higher decimal than the token
flow:
user clicks the "edit market price" CTA
try to edit the price of the token
expected: allow the user to set the price while ignoring the token decimal restrictions
actual: the token decimals are forced on the price input and prevent from setting accurate prices

https://github.com/bancorprotocol/carbon-app/assets/40430285/8f7f657b-e7b7-4663-adf9-83ca10f31faf

GrandSchtroumpf:
fix #1224 

With existing market price
image

Without existing market price
image


Costa769 (approved):

tiagofilipenunes (dismissed):
LGTM, just one comment on mobile.

- [x] On smaller width screens, the "Use Market" text goes outside of the input price box

image

PR #1276, opened 2024-06-18
Add special char to roman char map

tiagofilipenunes:
- Add ₮ and ₿ to special char map


src/utils/pairSearch.ts
```diff
@@ -64,6 +64,13 @@ export const createPairMaps = (
   return { pairMap, nameMap };
 };
 
+const specialCharToRomanMap: Record<string, string> = {
+  '₿': 'b',
+  '₮': 't',
+};
+const specialStringToRoman = (value: string) =>
+  value.replace(/./g, (char) => specialCharToRomanMap[char] || char);
+
```
GrandSchtroumpf, line 73: Personally I would have done it in a 
```suggestion
const replaceSpecialCharacters = (value: string) => {
  return value.replace(/(₿)|(₮)/g, (match, p1, p2) => {
    if (p1) return 'b';
    if (p2) return 't';
  });
}
```

src/utils/pairSearch.ts
```diff
@@ -79,9 +86,11 @@ export const searchPairKeys = (
   transformSlugExp: RegExp = pairSearchExp
 ) => {
   // Transform search into pair
-  const searchSlug = fromPairSearch(search, transformSlugExp);
+  const _searchSlug = fromPairSearch(search, transformSlugExp);
+  const searchSlug = specialStringToRoman(_searchSlug);
```
GrandSchtroumpf, line 90: I think it's better to include this replacement in the `fromPairSearch` & `createPairMaps`.

ashachaf (approved):

PR #1277, opened 2024-06-18
Issue #1263 - redirect to defaultTradePair on explorer page

### LINKED ISSUE #1263: Explore - update default view to pre select a token pair
Currently:
When a user reach the "explore" page, we require to select a token pair or wallet address in order to view any strategies

Updated:
on app -> pre select ETH/USDC
on sei -> pre select SEI/USDC

app, navigate to "explore" and we show the data available here: https://app.carbondefi.xyz/explore/token-pair/0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee_0xa0b86991c6218b36c1d19d4a2e9eb0ce3606eb48

tiagofilipenunes:
- Fix returning early in useGetPairStrategy which was resulting in isLoading becoming true before redirect happened
- Add redirect to explorerPage

fix #1263


src/libs/queries/sdk/strategy.ts
```diff
@@ -228,8 +228,7 @@ export const useGetPairStrategies = ({ token0, token1 }: PropsPair) => {
   return useQuery<Strategy[]>(
     QueryKey.strategiesByPair(token0, token1),
     async () => {
-      if (!token0 || !token1) return [];
-      const strategies = await carbonSDK.getStrategiesByPair(token0, token1);
+      const strategies = await carbonSDK.getStrategiesByPair(token0!, token1!);
       return await buildStrategiesHelper({
```
GrandSchtroumpf, line 232: I know it's not part of the PR, but you can remove the await here.

src/pages/explorer/index.tsx
```diff
@@ -9,9 +9,27 @@ import { StrategyProvider, useStrategyCtx } from 'hooks/useStrategies';
 import { ExplorerTabs } from 'components/explorer/ExplorerTabs';
 import { useEffect, useState } from 'react';
 import { explorerEvents } from 'services/events/explorerEvents';
+import { toPairSlug } from 'utils/pairSearch';
+import config from 'config';
 
 export const ExplorerPage = () => {
   const { slug, type } = useExplorerParams();
+  const navigate = useNavigate();
```
GrandSchtroumpf, line 17: Since you know the page you're on you can do that : 
```suggestion
  const navigate = useNavigate({ from: '/explore/$type/$slug' });
```

src/pages/explorer/index.tsx
```diff
@@ -9,9 +9,27 @@ import { StrategyProvider, useStrategyCtx } from 'hooks/useStrategies';
...
+
+  useEffect(() => {
+    if (slug || type === 'wallet') return;
```
GrandSchtroumpf, line 20: I think it's more readable like that. Also if we add a type in the future this would be the correct logic.
```suggestion
    if (slug || type !== 'token-pair') return;
```

GrandSchtroumpf (approved):

GrandSchtroumpf (approved):

src/pages/explorer/index.tsx
tiagofilipenunes: Agreed

PR #1280, opened 2024-06-19
[Overlapping] trim spread to 6 decimals

### LINKED ISSUE #1245: Overlapping - spread custom value only displays 2 decimals
## Current state
1. Custom spread input field accepts unlimited amount of digits
for this:
- user can input a value with 20 decimals with no restrictions

2. Custom spread input field, always shows 2 decimals points.
for this:
- if the user input 1.1111 -> the display would be 1.11
- if the user input 0.001->the display would be 0.00

## Update:
1. Custom spread input field should be limited to a max of 6 decimals.
for this:
- if the user inputs 1.123456789 -> the input will not allow to type a digit past 1.123456

2. Custom spread field will display more decimals (up to 6) of the custom spread following these rules

### Logic

The maximum number of decimals to be displayed in the custom spread fields-> 6 decimals

For every spread input value
- Select first 6 decimal digits of the input value
- Trim all "0" digits up to the first non 0 digit from right to left
- Display the result in the input field

### Example:
**User input-> display**
0.001000 -> 0.001
0.00101 -> 0.00101
0.00001111 -> 0.000011
0.100010 -> 0.10001
1.00000123->1.000001
1.1100001232-> 1.11

GrandSchtroumpf:
fix #1245 
image


src/components/strategies/overlapping/OverlappingSpread.tsx
```diff
@@ -91,7 +91,7 @@ export const OverlappingSpread: FC<Props> = (props) => {
         setSpread(defaultValue);
         e.target.value = '';
       } else {
-        e.target.value = Number(value).toFixed(2);
+        e.target.value = Number(Number(value).toFixed(6)).toString();
```
tiagofilipenunes, line 94:
```suggestion
        e.target.value = Number(value).toFixed(6).toString();
```

tiagofilipenunes (approved):

src/components/strategies/overlapping/OverlappingSpread.tsx
GrandSchtroumpf: It doesn't work, try in the console 
```
Number('0.001').toFixed(6).toString(); 
> "0.001000" 
```

ashachaf:
- [ ] input limitation
current:
when typing 0.1234567890123456789 
- the input allows user input to proceed past the 6th decimal
- when focus out of the input field, it trims the visible data to 6 decimals

updated:
when typing 0.1234567890123456789 
- limit the input field to only accept 6 decimals. do not allow input any value past the 6th decimal.

tiagofilipenunes (approved):

ashachaf (approved):

PR #1286, opened 2024-06-19
[Overlapping] Use correct market price

tiagofilipenunes:
# Market price for overlapping strategies

## Market price definition
**User defined market price** is defined if user fills the input.
**Calculated market price** is defined if strategy has budget
**External market price** is defined if both token has an entry in coingecko

## Market price order
We should always use the market price in this order of importance (if 1 is not defined we use 2, ...) : 

### Edit price
If price/spread has not changed : 
1. User defined market price
2. Calculated market price
3. External market price

If price/spread has changed :
1. User defined market price
2. External market price
3. User defined market price (with warning)

### Edit budget
1. User defined market price
2. Calculated market price
3. External market price (with warning)

If there is none of the market prices above is defined, we display the market price input to the user.


ashachaf:
- [ ] edit price - user is booted to main page
when trying to edit price for a newly created strategy, few seconds after the user arrive the edit prices page, it is booted back to the main strategy page. only happens on first try.
flow:
- create strategy 
- wait for it to show up on "my strategies" page
- click "edit prices"
expected: stay on edit prices page
actual: only on first try, get kicked to my strategy page

tiagofilipenunes (commented):
LGTM, would be ideal to have some e2e or unit tests with this change but they can be done separately.

Found this flow with a bug:
- [x] Start with an inactive strategy with tokens with unknown price -> edit price or renew -> recurring option is the default, change to overlapping. The form shows up as:

image

To get unknown prices just change the Carbon API url to a non-usable one

ashachaf:
- [ ] withdraw from strategy id 7486212072260646196194241363498900652154
try to withdraw 0.000000000000000004 GOKU
error:
update mutation failed Error: division-by-zero [ See: https://links.ethers.org/v5-errors-NUMERIC_FAULT-division-by-zero ] (fault="division-by-zero", operation="div", code=NUMERIC_FAULT, version=bignumber/5.7.0)

- [ ] CTA stays disabled
image
while editing strategy: 8507059173023461586584365185794205286538
- update market price
- select token to keep
- confirm checkbox
expect to see the cta enabled.

- [ ] edit market price trigger metamask popup
- edit prices
- change market price
actual: metamask popup shows up and ask the user to sign the transaction
expected: nothing to happen until i click the cta.

ashachaf (approved):

PR #1294, opened 2024-06-20
Improve multi-chain support locally and move env variables to config

tiagofilipenunes:
Please read the description to facilitate QA.

Proposed changes:
- Update to the way rpc url, headers, keys and .env variables are set:
  - Rpc url and headers have been moved to the config, with different development and production configs
  - There are now two types of rpc url and headers that can be defined. VITE_CHAIN_RPC_URL and config.rpc.url. If VITE_CHAIN_RPC_URL is not set, then config.rpc.url will be the one used for any network requests. For all cases, walletConnectRpc is used on wallet connection.
  - Remove VITE_CHAIN_RPC_URL - these env variables are already exposed in the network tab so no point in adding them. It also facilitates forks to set up their own development and production rpc urls. It was also not possible to set custom headers or key for the VITE_CHAIN_RPC_URL in case it was needed
  - Remove VITE_CHAIN_RPC_KEY as it was not used anymore
  - Note: Any fork will use the same rpc urls and headers set in the config. This is okay as the production ones are not expected to work (but can if they don't have a whitelist), and the development ones should be widely available or free rpc urls to facilitate running the app locally.

- The following .env variables were NOT moved to the config. Here are some discussion points about why not:
  - SDK
    - Variables: VITE_SDK_VERBOSITY and VITE_LEGACY_TRADE_BY_SOURCE_RANGE
      - Discussion
        - Makes it easy to set via deployments, obfuscating these options is good as they are optional and when required, the user will be familiar with them
        - SDK also expects env variables so we can keep a consistent behaviour with these flags in case some forks also run the SDK independently or check the documentation and are confused
        - Conclusion: No real benefit to adding them to config
  - Sentry
     - Variables: SENTRY_ORG, SENTRY_PROJECT, and SENTRY_AUTH_TOKEN
       - Discussion
         - Might be useful to add it here to make it obvious for projects wanting to use Sentry
           - Can add it to the readme instead
         - If someone forks the code, they will also port these variables and even though Sentry filters the origin domain, it's bad practice to let forks use these params as they are deployment-specific
         - It's much easier to change these settings via env variables without having to do a PR
         - Rotating keys would mean creating a PR which means better visibility for contributors
         - Conclusion: More points in favour of not adding to config
  - E2E
    - Variables: CI and TENDERLY_ACCESS_KEY
    - Conclusion: Do not move these to config as they are set by E2E CI and being able to override CI locally to produce screenshots is useful

- Facilitate local development in multiple chains but keep build process
  - start:network command as been added to the package.json with instruction in the readme on how to add it for forks
  - Local storage is partitioned by using the network when setting the keys as `[APP_ID, NETWORK, APP_VERSION, key].join('-')`
  - Clean-up and migration process of old locale storage keys created and instructions added to the readme
  - VITE_NETWORK can still be used when running yarn start and not using yarn start:network options. This variable must be set when building
- Update github CI workflows:
  - Update `build_test.yaml` to remove unused env variables and update all actions to ones that default to using Node v20
  - Update `e2e.yaml` to remove VITE_CHAIN_RPC_URL as it is no longer used
  - Update test.yaml actions to ones that default to using Node v20


src/hooks/useCarbonInit.ts
```diff
@@ -36,6 +36,17 @@ const getTokenDecimalMap = () => {
   );
 };
 
+const removeOldV1Keys = () => {
+  Object.keys(localStorage).forEach((key) => {
+    if (
+      key.match(/carbon-v1(\.1)?-([a-zA-Z0-9]+-?)+/) &&
```
zavelevsky, line 42: why not simply keys that start with it? What's the reason for the extra checks?

tiagofilipenunes: I moved the network in the prefix to the order of [APP_NAME, NETWORK, APP_VERSION] in a previous commit, exactly to simplify this check but forgot to change it. Done.

README.md
```diff
@@ -43,7 +42,15 @@ yarn install
 yarn start
 ```
 
-Open [http://localhost:3000](http://localhost:3000) to view it in the browser.
+You can choose the network to run the app by setting the `VITE_NETWORK` variable in the `.env` file. The default network is `ethereum`.
+
+Alternatively, if defined in the in the [package.json](package.json) file, you can select the network to use by running `yarn:network` where `network` is one of the networks defined in the [config](src/config/index.ts) and the `yarn:network` command is defined in [package.json](package.json) file as follows:
+
+```bash
+"start:network": "export VITE_NETWORK=network && vite",
```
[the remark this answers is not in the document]
tiagofilipenunes: I've added a script that uses node's `spawn` to start vite with the specified network.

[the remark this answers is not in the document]
tiagofilipenunes: Instead used [cross-env](https://github.com/kentcdodds/cross-env)

src/hooks/useCarbonInit.ts
tiagofilipenunes: Added key migration and removal logic

src/utils/managedLocalStorage.ts
```diff
@@ -49,4 +53,16 @@ export class ManagedLocalStorage<T> {
     const formattedId = this.keyFormatter(key);
     localStorage.removeItem(formattedId);
   };
+
+  migrateItems = () => {
+    if (!this.migrations?.length) return;
+    this.migrations.forEach(({ matcher, action }) => {
+      Object.keys(localStorage).forEach((oldFormattedKey) => {
+        const key = matcher(oldFormattedKey);
```
zavelevsky, line 61: You have 2 main problems here:
1. The API is too chatty. it can be as simple as migration.migrate(key) - why should the caller care about whether or not there was a match and formatting etc.
2. A migration step knows how to convert legacy format to the next format - and go through all the steps. Your reliance here on the up to date key formatter is wrong - as it will change over time. Migration code is something that stays fix, and it will forever change "A" to "B", and the next one will change "B" to "C", and some time later when C is outdated we will add a migration for "C" to "D".

Talk to me if something isn't clear

src/services/localeStorage/index.ts
```diff
@@ -46,6 +51,36 @@ interface LocalStorageSchema {
...
 }
 
-export const lsService = new ManagedLocalStorage<LocalStorageSchema>((key) =>
-  [APP_ID, APP_VERSION, key].join('-')
+// ************************** /
+// LOCAL STORAGE MIGRATION AND CLEAN_UP
+// ************************** /
+
+/**
+ * An array of migration objects, each designed to handle specific versions of stored items.
```
zavelevsky, line 59: emphasize that order matters and that each migration step is applied after the other, so that even if someone open the app after a year and 20 upgrades, there will be a series of migration steps that take him all the way to now

src/hooks/useCarbonInit.ts
```diff
@@ -130,6 +130,7 @@ export const useCarbonInit = () => {
       if (isBlocked && !lsService.getItem('hasSeenRestrictedCountryModal')) {
         openModal('restrictedCountry', undefined);
       }
+      lsService.migrateItems();
```
zavelevsky, line 133: migration should happen prior to any local storage access.

src/utils/migratorLocalStorage.ts
```diff
@@ -0,0 +1,104 @@
...
+export class MigratorLocalStorage {
+  private readonly migrations: Migration[] = [];
+  constructor(migrations?: Migration[]) {
+    if (migrations) this.migrations = migrations;
+  }
+
+  migrateItem = (prevFormattedKey: string) => {
+    this.migrations.forEach(
+      ({ action, prevKeyExtractor, nextKeyFormatter }) => {
+        const key = prevKeyExtractor(prevFormattedKey);
```
zavelevsky, line 90: why isn't this simply

```ts
this.migrations.forEach(migration => migration.migrate(prevFormattedKey));
```

and let the migration internally do all this logic of lines 90-93?

src/hooks/useCarbonInit.ts
tiagofilipenunes: Moved to the top of the init function

src/utils/migratorLocalStorage.ts
tiagofilipenunes: Updated to this pattern

PR #1301, opened 2024-06-24
[Wallet] Add tailwind wallet

tiagofilipenunes:


ashachaf:
- [ ] auto connect doesnt work
1. connect wallet
2. refresh page
expected: auto connect
actual: app asks to reconnect the wallet

- [ ] reject is not picked up by app
1. connect wallet
2. create a trade on the app
3. open wallet dialogue and reject it
expected: CTA to pick up that user rejected and return to the state of "buy"
actual: CTA stays on loading mode waiting for user response
image

Issue #1302, opened 2024-06-24
[BUG] No approval re-request for consecutive trades using limited approval


tiagofilipenunes:
How to reproduce:
1. Trade an amount of an ERC-20 on the trade page
2. When showed the approval modal, approve 'Limited' (default on Sei)
3. Swap 
4. Once 3) confirms, attempt to swap the same ERC-20 and same amount again
5. The approval modal does not show and the transaction fails

https://github.com/bancorprotocol/carbon-app/assets/66181828/1efbc1b7-613e-41d7-9106-88f1a45de19c

Reproduced on Chrome 126.0.6478.63 / MacOS 14.5

PR #1303, opened 2024-06-24
Issue #1302 - Request approval again after successful trade

tiagofilipenunes:
- Set get approval query invalidation after transaction success
- Improve e2e trade page test coverage


GrandSchtroumpf (approved):
Crazy that the onSuccess happened before the tx.wait().
Good catch !

Issue #485, opened 2023-03-28
Delay between successful transaction notification and strategies page update 


Costa769:
Issue:

When the user adds a new strategy and he sees the transaction success notification , there is a small delay before the strategies page is updated and the new strategy becomes visible.

Suggestions:

1. Get the transaction status from the SDK instead of directly from the blockchain
2. Once the successful transaction notification is displayed, add a new strategy just with the Carbon icon animation until the strategy details are received from the SDK 

Mockup:
<img width="911" alt="image" src="https://user-images.githubusercontent.com/127974452/228216123-0b0ed490-d3f9-444f-ad00-683f4a008a60.png">

ashachaf:
@Costa769 is this still relevant?

tiagofilipenunes:
Issue #1302 and #1304 address this problem and are done, worth checking again if this is still relevant.

PR #1304, opened 2024-06-24
[My Strategies] Add update indicator

### LINKED ISSUE #1300: [My Strategies] Updating state
When creating/editing a strategy we arrive on "My Strategies" page and the updates only arrives after some seconds.
In the meantime there is no indication that the state is stale. This has two effects : 
- UX: user has any indication on what's happening which can be stressful, especially if transaction involves large budgets.
- E2E: tests don't know when the state is updated, which leads to flaky tests.

Solution : 
Display a fetching indication at the top of the strategy list.
- Since the query is on all strategies of the user, we don't know which strategy is loading, so we need a global loading indication.
- Ideally this should not create a layout shifting, so we need to keep the space available all the time.
- This should happen only on fetching a new state, and **not on initial load**.

It could look like this : 

https://github.com/bancorprotocol/carbon-app/assets/8143464/49f75ef0-51e7-47d7-b1bd-dfc963d166ee

GrandSchtroumpf:
fix #1300 

Add an indication that the state is updating in My Strategies page.
This happens only for update, not for the initial loading.

https://github.com/bancorprotocol/carbon-app/assets/8143464/e76bf925-99a2-4fe0-86a9-d91eac0e959c


tiagofilipenunes (approved):
LGTM, hope this improves e2e test stability

Costa769:
- [ ] On the Ethereum app, the indicator is not displayed; despite making several deposits and withdrawals, the indicator does not appear.
- [ ] On the SEI app, the indicator is visible for only half a second. Perhaps we can extend its appearance by 1-2 seconds.

tiagofilipenunes (approved):
LGTM

tiagofilipenunes:
fix #485 cc @Costa769

GrandSchtroumpf:
@tiagofilipenunes it doesn't actually fix the issue since the indicator is now totally transparent. At this point, it's only for E2E.

PR #1311, opened 2024-06-25
[Withdraw Disposable] Display inactive budget for disposable withdraw

### LINKED ISSUE #1310: Withdraw - disposable strategy flow
when withdrawing from a disposable strategy we need to check the budget or the inactive order and display it if not zero.

flow:
- withdraw disposable strategy
- check: IF inactive order budget != 0, expose it as we do for recurring strategy

GrandSchtroumpf:
fix #1310 

Withdraw with both budget
image

Withdraw with active only 
image

Withdraw with inactive filled & active empty
image


ashachaf (approved):

tiagofilipenunes (approved):
LGTM, just one comment.

- When I pause a strategy and deposit to or withdraw from it, I'm directed to the budget/recurring page instead of budget/disposable. The latter is now showing all the non-zero balances as opposed to production if I go to the link directly. But through the app, I couldn't find the flow described in the issue #1310

PR #1314, opened 2024-06-26
Add dragonswap tokenlist

tiagofilipenunes:


ashachaf (approved):

src/config/sei/utils.ts
```diff
@@ -59,3 +38,49 @@ export const tokenListParser =
     };
     return parsedData;
   };
+
+const getGitFolderContent = async (
+  gitRepoInfoApi: string
+): Promise<unknown> => {
```
GrandSchtroumpf, line 44: no need to hard type the result if it's unknown, it's already the result of response.json()

GrandSchtroumpf (approved):

PR #1317, opened 2024-06-27
[Overlapping] mark none overlapping strategy as touched

GrandSchtroumpf:
Fix bug when switching from a recurring to an overlapping strategy and only changing the budget
- Create a recurring strategy
- Edit Price
- Click on "Concentrated"
- Try to update budget

Expected: both anchor are enabled
Current: both anchor are disabled


tiagofilipenunes (approved):

ashachaf:
- [ ] switch from concentrated to recurring is disabled

tiagofilipenunes (approved):
LGTM

PR #1321, opened 2024-06-28
[Overlapping] Graph uses external market price

### LINKED ISSUE #1320: [Overlapping] Graph should display the calculated values of the order & external market price
Currently the graph displays the market price used for the calculation, while it should display the external market price if used.

1. Display user market price if defined
2. Display external market price if defined
3. Else display calculated market price

The marginal prices should reflect the marginal price the user will get when he click on the CTA
- If not touched: use existing marginal price
- If touched: use calculated marginal price

GrandSchtroumpf:
fix #1320 

Strategy with external market price:
image

Strategy without external market price
image

Note: The text is "Geometric mean", while it's not exactly that when strategy is above/below market. Maybe use "Calculated market price".

Strategy with user market price
image

Switching from untouched to touch changes the marginal price: 
https://github.com/bancorprotocol/carbon-app/assets/8143464/ce548f8d-c799-4b85-bec0-4783b09133e3


ashachaf:
- [ ] edit price - do not default into price field when price is available
flow:
1. click edit price
2. IF external price is available, do not default into the price field (because it force to show the last bit of the price which is usually a number that doesnt make sense)

https://github.com/bancorprotocol/carbon-app/assets/40430285/2ddb9d14-cc1a-466a-bbe1-5b931647f31b

- [ ] external price - decimal trim
currently, we show 82 decimals.
updated, show the following:
IF (first 6 decimals are 0) AND (price < 1), show the 6 decimals after the first non zero digit
ELSE, show 6 decimals

- [ ] edit market price when there is no external price
currently warning:
Warning, your market price will be used in the strategy creation flow and calculations.

updated warning:
Warning, the defined market price will be used for strategy budget distribution
 
image

- [ ] edit strategy with no external price
flow:
1. edit a strategy with no external price
2. show the price input modal
image
3. in the place where we typically show "use market" and allow users to click and use the external market price, replace it with "use strategy" and allow users to click and use the strategy geomean(marginal.sell, marginal.buy)
4. add tool tip when mouse over "use strategy" with this text: "This price is the geometric mean of the strategy buy and sell marginal prices."

Costa769:
- [ ] Updating prices on the graph

flow:

1. Find concentrated strategy without market price
2. Click edit prices
3. Chage the prices via the graph without providing the market price

Expected:
User can't update the prices

Actual:
User can change the prices of the strategy

- [ ] Deposit/withdraw from a strategy without market price

flow:

1. Find concentrated strategy without market price
2. Click edit prices
3. Select one of the tokens and click on edit budget without providing the market price
4. Input deposit amount
5. Click the checkbox

Expected:
CTA becomes enabled

Actual:
CTA is disabled and becomes enabled only after user sets the market price

ashachaf:
- [ ] Using external market cap for deposit/withdraw incorrectly
Flow:
- edit concentrated strategy with market price
- click to withdraw some tokens (no other changes)
expected: withdraw from the strategy according to the on-chain marginal price
actual: the check is using the current market price and tried to re-adjust the strategy around it
image

ashachaf:
- [ ]text change
current: 
{token} market price is missing. To continue the prices of the strategy, enter its market price below.

updated:
{token} market price is missing. Please provide the market price to enable price editing.

Costa769:
- [ ] Difference from market price indicators

image

Current: We always show difference from market price indicators, wheter there is a market price or not
Updated: Show difference from market price indicators only when there is an external market price or market price that was provided by the user

ashachaf (approved):

src/components/strategies/create/CreateOverlapping.tsx
```diff
@@ -42,7 +42,8 @@ const url = '/strategies/create/overlapping';
 export const CreateOverlapping: FC<Props> = (props) => {
   const { base, quote, order0, order1, marketPrice, spread } = props;
   const navigate = useNavigate({ from: url });
-  const { anchor, budget } = useSearch({ from: url });
+  const search = useSearch({ from: url });
+  const { anchor, budget } = search;
```
zavelevsky, line 46: why not also extract marketPrice here?

GrandSchtroumpf: There are multiple type of market price, the one from the search is the user market price and might not be there. We use different market price depending on the situation, and this calculation is done at parent component level.

GrandSchtroumpf:
Manually tested with :

1. Create an overlapping strategy with tokens with market price
2. Trade against it
3. Go to edit price, the graph should display external market price, but marginal price is the current one from the strategy
4. Try to withdraw without change price or spread, it should use the calculated market price
5. Withdraw

1. Create an overlapping strategy with tokens without market price
2. Trade against it
3. Go to edit price, the graph should be disabled (you cannot change the price), and the price section is hidden
4. Withdraw -> it should use calculated market price
5. Go back to edit price, enter a market price in the input field -> the graph should be enabled and price is available
6. Withdraw -> it should use the market price you filled in the input

zavelevsky (approved):

PR #1323, opened 2024-07-01
[Test] Setup integration tests

tiagofilipenunes:
Simple POC to setup the integration library before starting to write integration tests.

Changes made:
- Add jest-dom eslint plugin
- Update testing-library packages
- Update vitest and vite
- Added examples of vitest.mock and vitest.hoisted - as mock calls are hoisted, you can use vitest.hoisted to run code before the mocks
- vite-plugin-rewrite-all plugin was removed as part of updating vite as it is no longer used and has been deprecated
- Added two customRenders - both with some providers added and one with the router added as well
- Added wagmi wallet mock

Notes:
- Currently tests are running with mode flag production because otherwise there's an error when loading the config. Alternatively, the config file can be mocked and this flag removed
- Similar as above for the compiled abis - we currently compiled them before the tests and alternatively they can be mocked


src/components/common/WarningMessageWithIcon.test.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <Warning htmlFor="" message="" className="" isError={true}>
+      Example warning
+    </Warning>
+  );
+};
+
+test('Test warning component', async () => {
+  // ARRANGE
+  await renderWithRouter({ component });
```
GrandSchtroumpf, line 21: Feels like this is the kind of component that do not need the router. I think it's nice if we can have an example of a page (need the router) and a reusable component (don't need the router).

src/libs/testing-library/index.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+export const renderWithRouter = async (params: RouterRenderParams) => {
+  const customRouter = await loadRouter(params);
+  return {
+    container: customRender(<RouterProvider router={customRouter} />),
+    router: customRouter,
+  };
+};
+
+export * from '@testing-library/react';
```
GrandSchtroumpf, line 37: Any reason to reexport everything ?

src/libs/testing-library/setup.ts
```diff
@@ -0,0 +1,71 @@
...
+        isError: false,
+        error: null,
+        setImportedTokens: () => {},
+      };
+    },
+  };
+});
+
+// MOCK CARBON SDK
+enum MarginalPriceOptions {
```
GrandSchtroumpf, line 29: I think you can hardcode it directly in the mock, no need to create an enum

src/libs/testing-library/utils.ts
```diff
@@ -0,0 +1,47 @@
...
+    symbol: 'ETH',
+  },
+  {
+    address: debugTokens.USDC,
+    decimals: 6,
+    symbol: 'USDC',
+  },
+];
+
+export const loadRouter = async ({
```
GrandSchtroumpf, line 24: Could you add some comments inside this function to understand the process ?

vite.config.ts
```diff
@@ -3,20 +3,14 @@ import react from '@vitejs/plugin-react';
...
   // Set the third parameter to '' to load all env regardless of the `VITE_` prefix.
   const env = loadEnv(mode, process.cwd(), '');
 
-  const plugins: PluginOption[] = [
-    react(),
-    viteTsconfigPaths(),
-    svgrPlugin(),
-    pluginRewriteAll(),
-  ];
+  const plugins: PluginOption[] = [react(), viteTsconfigPaths(), svgrPlugin()];
```
GrandSchtroumpf, line 13: Do you know what `pluginRewriteAll` was doing ? Maybe @pingustar you know ?

src/components/common/WarningMessageWithIcon.test.tsx
tiagofilipenunes: You're right, this was just while I was testing out the router, We already have MarketPriceIndication as an example without the router, so I'll remove this file

src/libs/testing-library/index.tsx
tiagofilipenunes: To force imports to come from libs/testing-library as per their docs

https://testing-library.com/docs/react-testing-library/setup

src/libs/testing-library/setup.ts
tiagofilipenunes: You're right, removed it

vite.config.ts
tiagofilipenunes: I added it - it's used on dev server to not rewrite the path if it includes a dot '.'. This was useful for explorer search by ens as all the links would have a dot in the url and we couldn't test it locally.

Package here: https://github.com/ivesia/vite-plugin-rewrite-all

This is no longer required, check here:

https://github.com/vitejs/vite/pull/14142

src/libs/testing-library/utils.ts
tiagofilipenunes: Added jsdoc

GrandSchtroumpf (approved):
Great work !

src/components/strategies/marketPriceIndication/MarketPriceIndication.test.tsx
```diff
@@ -23,7 +15,7 @@ describe('MarketPriceIndication', () => {
     );
 
     const indicationText = screen.getByText('5.00% above market');
-    expect(indicationText).toBeTruthy();
+    expect(indicationText).toBeInTheDocument();
```
zavelevsky, line 18: this is probably a redundant assertion because if it wasn't present the previous line would throw:
https://testing-library.com/docs/queries/about#types-of-queries

src/pages/strategies/create/recurring.test.tsx
```diff
@@ -0,0 +1,33 @@
+import { test, expect, afterEach } from 'vitest';
+import { cleanup, renderWithRouter, screen } from 'libs/testing-library';
+import { debugTokens } from '../../../../e2e/utils/types';
+import { CreateRecurringStrategyPage } from './recurring';
+
+const basePath = '/strategies/create/recurring';
+afterEach(async () => {
+  cleanup(); // Clear the screen after each test
```
zavelevsky, line 8: no need to call cleanup

zavelevsky (changes requested):

src/pages/strategies/create/recurring.test.tsx
tiagofilipenunes: Vitest doesn't automatically call clean-up. 

https://github.com/vitest-dev/vitest/issues/1430

Tried changing the isolate and thread options but still ran into issues where rendering was adding elements without cleaning up between tests

zavelevsky (approved):

PR #1330, opened 2024-07-04
testing the CI running 4 workers

zavelevsky:


tiagofilipenunes (approved):
LGTM, just a suggestion to modify the comment about no. of workers from a fixed number (currently [max cores that GH provides for actions](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories)) to 100% of the cores.

PR #1334, opened 2024-07-05
Add create overlapping integration tests

tiagofilipenunes:
Add overlapping tests:
- should open and close chart
- should populate form with search params and user market price defined in url
- should populate form and search params with unavailable external price after setting user market price
- should populate form and search params with available external market price
- should set default spread with spread unset in the search params
- should populate form with user market price below min price
- should populate form with user market price above max price
- should populate form without min and max defined and with user market price
- should populate form with invalid range in search (min>max
- should check form touched: after changing min range
- should check form touched: after changing max range
- should check form touched: after changing spread
- should check form touched: after setting user price


src/pages/strategies/create/overlapping.test.tsx
```diff
@@ -0,0 +1,70 @@
+import { test, expect, afterEach, vitest } from 'vitest';
+import { renderWithRouter, screen } from 'libs/testing-library';
+import { debugTokens } from '../../../../e2e/utils/types';
+import { CreateOverlappingStrategyPage } from './overlapping';
+import userEvent from '@testing-library/user-event';
+
+vitest.mock('hooks/useMarketPrice', () => {
```
GrandSchtroumpf, line 7: wouldn't it be easier to mock the endpoint directly like we do in playwright ?

src/pages/strategies/create/overlapping.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+
+const basePath = '/strategies/create/overlapping';
+afterEach(async () => {
+  window.history.replaceState(null, 'root', '/');
+});
+
+test('run overlapping form with deeplink and user market price', async () => {
+  // Initialize user and set params
+  const user = userEvent.setup();
+  const search = {
```
GrandSchtroumpf, line 23: Ideally you would use typescript interface to make sure we update there if there is a change in the interface in the future

src/pages/strategies/create/overlapping.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+    basePath,
+    search,
+  });
+
+  // Check search params
+  expect(router.state.location.pathname).toBe(basePath);
+  expect(router.state.location.search).toStrictEqual(search);
+
+  // Close price chart
+  await user.click(screen.getByText('Close Chart'));
```
GrandSchtroumpf, line 46: As discussed in the channel, I don't like using text or role query. I think `data-testid` is more consistent. Also it ensure separation of concern : 
- query with testid make sure you have to element
- check role or text with except to make it explicit that you want to verify this information

src/libs/testing-library/utils/routing.ts
```diff
@@ -1,25 +1,26 @@
...
-    symbol: 'ETH',
-  },
-  {
-    address: debugTokens.USDC,
-    decimals: 6,
-    symbol: 'USDC',
-  },
-];
+const encodeValue = (value: string | number | symbol) => {
+  if (typeof value == 'string' && isAddress(value)) return `${String(value)}`;
```
GrandSchtroumpf, line 12: Why wrap the value with `String` ? the back quote should do it no ?

zavelevsky (approved):

src/libs/testing-library/utils/mock.ts
```diff
@@ -0,0 +1,120 @@
+import { debugTokens } from '../../../../e2e/utils/types';
+import { RequestHandler } from 'msw';
+import { setupServer } from 'msw/node';
+import { HttpResponse, http } from 'msw';
+import tokenListsMock from '../../../../e2e/mocks/tokenLists.json';
```
GrandSchtroumpf, line 5: Suggestion: If we use E2E files, maybe we could move the testing logic into a root folder : 
`tests/e2e`
`tests/integration`

GrandSchtroumpf (approved):

src/libs/testing-library/utils/routing.ts
tiagofilipenunes: That's right - removed

src/libs/testing-library/utils/mock.ts
tiagofilipenunes: Not doing for now as got a lot of issues with imports in *.test.* files as well as using vi.mock when moving the testing-library to tests/integration

PR #1335, opened 2024-07-05
[E2E] disable button when approval is loading

GrandSchtroumpf:
E2E try to submit a transaction before approval returns: disable button while approval is pending


tiagofilipenunes (approved):
LGTM

Issue #1292, opened 2024-06-20
[Activity] Filtering in activity with special characters doesn't work


tiagofilipenunes:
Related to #1251, the activity filter options by strategies and by pair to not translate the symbol ₮ as T. 

https://github.com/bancorprotocol/carbon-app/assets/40430285/e6968e70-f095-4455-a58b-2934f9e76827

PR #1340, opened 2024-07-10
in activity pair filter we now support special characters

zavelevsky:
fixes: #1292 

Every option in the pair combo is now compared as-is but also after converting any special characters (right now just ₿ -> b and ₮ -> t)


tiagofilipenunes (approved):
LGTM - good catch on the case-sensitive search. 

Tried locally with USDT, USD₮, usdt, uSdT, usd₮, usdc on the USD₮/SEI explore activity and worked well.

ashachaf (approved):

PR #1348, opened 2024-07-13
last used pair across all pages

### LINKED ISSUE #1331: Token pair default selection
## Overview

Currently, on the trade page, we apply a process in which the token pair is pre selected using these rules:
- First visit, default pair
- Any future visit, last user selected pair

The goal of this issue is to expand this functionality beyond the "trade" page and across all pages.
expected behavior: once the user select tkn1/tkn2 pair on any of the pages, this will be used as the default pair on all pages.

## Current Implementation
Trade page is currently using this logic. 

## Enhanced implementation
Perform the same check applied on the trade page, on every other page in the app and apply it on the token pair selection.

- Explore page -> apply the last pair value into the search query and display the information accordingly
- Simulate page -> apply the last pair value into the token selection (base/quote) and fetch the information for them
- Create strategy flow -> step 1 in this process requires the user to manually select base and then quote tokens. the expected behavior is that we apply the last pair value on page load.

current "create new strategy" flow:
https://github.com/bancorprotocol/carbon-app/assets/40430285/a0e95108-4fd3-465c-935b-6261014ddc02

updated "create new strategy" flow:
https://github.com/bancorprotocol/carbon-app/assets/40430285/c5d432ec-6433-4924-b8e5-e6414303cb5c

zavelevsky:
fixes #1331 

When visiting the following pages: trade, explore (by pair), simulate, create strategy - the selected pair becomes the default pair.

Note - the current behavior is buggy.
If you start to create a strategy and select a pair that doesn't exist yet it becomes the default pair, even if you don't complete the strategy creation - ending up with a bad "empty" experience in other pages.
It can also be more problematic with imported tokens etc.

I personally think that strategy creation shouldn't dictate the pair - or it should only take place once the strategy was created - or only if the pair already exists.


ashachaf (approved):

ashachaf:
- [ ] explore page cannot handle tokens with no pair
flow:
1. go to create strategy
2. import this token 0x809FF4801aA5bDb33045d1fEC810D082490D63a4
3. go to trade -> it identify there is no data for the token pair and indicate "no data"
4. go to explore
expected: identify there is no data for the token pair and indicate "no data" as we do for a pair with no strategies ([here](https://feature-default-pair-1331.carbon-app-sei.pages.dev/explore/token-pair/0x24e4d0af5a6b1c3bd336dd60deefa64c704b19b7_0xeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeeee))
actual: there is no token indication in the search url and the page is stuck on endless loading.

src/pages/explorer/index.tsx
```diff
@@ -10,17 +15,24 @@ import { ExplorerTabs } from 'components/explorer/ExplorerTabs';
...
+import { lsService } from 'services/localeStorage';
 
 export const ExplorerPage = () => {
   const { slug, type } = useExplorerParams();
   const navigate = useNavigate({ from: '/explore/$type/$slug' });
 
   useEffect(() => {
+    if (slug && type === 'token-pair') {
+      // Set last visited pair
+      const [base, quote] = slug.split('_');
```
tiagofilipenunes, line 27: Consider getting the baseToken and quoteToken from getTokenById like so and check they exist before setting the tradePair in local storage in case of wrong address pair

```typescript
      const [base, quote] = slug.split('_');
      const baseToken = getTokenById(base);
      const quoteToken = getTokenById(quote);
      if (!baseToken || !quoteToken) return;
```

src/pages/simulator/index.tsx
```diff
@@ -20,6 +22,15 @@ export const SimulatorPage = () => {
     start: searchState.start,
     end: searchState.end,
   });
+
+  useEffect(() => {
+    if (!searchState.baseToken || !searchState.quoteToken) return;
```
tiagofilipenunes, line 27: Same as above, something like this

```typescript
      const base = getTokenById(earchState.baseToken);
      const quote = getTokenById(searchState.quoteToken);
      if (!base || !quote) return;
```

tiagofilipenunes (changes requested):
Just two suggested changes for when the pair doesn't exist in the simulate and explore tab. The addresses may be present in the search but they can be invalid, in which case let's not set the tradePair

ashachaf (approved):

ashachaf (approved):

tiagofilipenunes (approved):

PR #1350, opened 2024-07-15
Remove manual definition of appIcon in metamask and coinbase connector

tiagofilipenunes:
Both metamask and coinbase wallets get the app icon from the favicon, so there is no need to set it manually. Further, this warning was being shown for the metamask connector:

```
Invalid dappMetadata.iconUrl: URL must start with http:// or https://
```

- Coinbase SDK gets favicon, no need to set the icon manually: https://github.com/coinbase/coinbase-wallet-sdk/blob/ebd73f23f92326f7de919f0927144f05a375a480/packages/wallet-sdk/src/CoinbaseWalletSDK.ts#L24

- Metamask SDK also gets favicon, no need to set the icon manually: https://github.com/MetaMask/metamask-sdk/blob/7ca8a7f7528d8498a07bf13cbda8c0e02cc0d90e/packages/sdk/src/services/MetaMaskSDK/InitializerManager/setupDappMetadata.ts#L61


zavelevsky (approved):
Approved

PR #1357, opened 2024-07-16
[Overlapping] Enable undercut

### LINKED ISSUE #1325: Strategy duplicate - expose undercut option for concentrated strategy
current:
when you duplicate a strategy:
IF strategy is disposable or recurring, display a popup asking if you would like to copy as is or undercut.
where undercut update the prices of the strategy by 0.1% closer together.

image

```
example:
existing strategy is selling tokens at $1500
undercut will set the price to be $1500*0.99% = $1485
```

updated:
add handling for concentrated strategy.

IF strategy is concentrated (using the same check we do for edit prices etc), display the same popup asking to copy as is or undercut
where undercut reduce the fee tier (spread) by 0.1%.

```
example:
existing strategy between usdc/usdt with price range of 0.999 to 1.001 and a spread of 0.01%
undercut will set the price the same (0.999 to 1.001) and update the spread to be 0.01*0.99 = 0.0099%
```

GrandSchtroumpf:
fix #1325

Enable undercut for overlapping strategy
Reduce spread by 0.1%.

Test cases : 
1. Create an overlapping strategy with spread of 1%
2. Menu -> Duplicate -> Undercut 
3. Spread is now 0.99%

This work if strategy :
- has no budget nor market price
- is above/below market price
- are built out of a recurring strategy (default to 0.05*0.99)


tiagofilipenunes (approved):
LGTM

Only thing to note is if the overlapping strategy has a defined budget, this budget (and anchor) won't be populated when undercutting, which is the same behaviour as overlapping duplicate but different from recurring/disposable undercut/duplicate.

ashachaf:
- [ ] remove duplicate option from strategies with no prices
IF order0.buyLow AND order0.sellHigh AND order1.buyLow AND order1.sellHigh = 0, remove the duplicate option from the menu

- [ ] duplicating recurring strategy opens concentrated creation
https://github.com/user-attachments/assets/23310a69-ceaa-4f04-93a7-ee9b8e1db32e
https://github.com/user-attachments/assets/2b4734b4-4d54-4d78-b973-0ed2c841e2fd

src/libs/modals/modals/ModalDuplicateStrategy/ModalDuplicateStrategy.tsx
```diff
@@ -7,22 +7,50 @@ import { ModalFC } from 'libs/modals/modals.types';
...
+import { StrategyType } from 'libs/routing';
 
 export type ModalDuplicateStrategyData = {
   strategy: Strategy;
 };
 
+const getStrategyType = (strategy: Strategy): StrategyType => {
+  const isBuyEmpty = !+strategy.order0.endRate;
+  const isSellEmpty = !+strategy.order1.endRate;
+  if (isBuyEmpty || isSellEmpty) return 'disposable';
```
tiagofilipenunes, line 22: Could you use `isDisposableStrategy` from `src/components/strategies/common/utils.ts` here?

tiagofilipenunes (changes requested):
LGTM just one change request for the e2e test

1. Currently, the create/duplicate/undercut tests all override the create/form.png screenshot. Can you change the submit method of CreateStrategyDriver to something like this

```typescript
  async submit(type: StrategyCase) {
    const btn = this.page.getByText('Create Strategy');
    if (shouldTakeScreenshot) {
      const mainMenu = new MainMenuDriver(this.page);
      await mainMenu.hide();
      await waitTooltipsClose(this.page);
      const form = this.getForm();
      const path = screenshotPath(this.testCase, type, 'form');
      await screenshot(form, path);
      await mainMenu.show();
    }
    try {
      await waitFor(this.page, 'approve-warnings', 2_000);
      if (await this.page.isVisible('[data-testid=approve-warnings]')) {
        await this.page.getByTestId('approve-warnings').click();
      }
      await btn.click();
    } catch {
      await btn.click({ timeout: 1_000 });
    }
  }
```

and then set they type to the strategy type.

tiagofilipenunes (approved):

PR #1360, opened 2024-07-16
[Strategy Card] Truncate number to 6 decimal if range < 1

### LINKED ISSUE #1273: Strategy overview - increase decimals for the strategy grid and market price indicator
## Overview

In the current implementation on the strategy card grid and in the price tooltip, we show 6 decimal places if the price is less than 1 and 2 decimal places if the price is 1 or greater. This creates a precision issue, as some prices display only 2 decimal places, while showing the differences between these prices requires exposing more decimal places.

#### Example
Strategy overview card
image

Price tooltip
image

Therefore in cases when the price grid is very narrow and requires the 3rd, 4th, 5th or 6th decimal to indicate difference, we would like to increase decimal visibility and show higher precision prices.

## Implementation:

Note that the below logic should be applied to the following:
- price values on the grid 
- buy and sell price tooltips

### Steps to Determine Decimal Places:
NOTE: we should 
IF maxSell - minBuy >1, 
[True] display 2 decimals. //this will allow gaps of more than 1 unit to display 2 decimals which is sufficient
[False] else, display 6 decimal after the last non zero value //this will allow to display small gaps with high precision while applying the subscript if needed.

```
Examples
Given the strategy overview card grid and sell price tooltip examples in the overview section
maxSell: 1.000500000000000672293211306007000873418655194639054916625613448333785358652768944304238066464084981
trimMaxSell = apply subscript rule on the number: 1.000500

minBuy: 0.999499999999995715780598558949538517518149610577459147388312832305246047326363623142242431640625
trimMinBuy = apply subscript rule on the number: 0.999499

Strategy overview card current display:
Price points: 0.999531, 0.999844, 1.00, 1.00 
Market price: 1.00

After applying the new logic: 
IF trimMaxSell - trimMinBuy > 1 -> 1.000500 - 0.999499 > 1 -> 0.001001 > 1 -> FALSE
- False means display 6 decimals
- Price points: 0.999531, 0.999844, 1.000157, 1.000470 
- Market price: 1.000287
- Tool tip Min price: 0.9996
- Tool tip Max price: 1.000515
- Marginal price: 0.999601
```

```
example 2:
MaxSell: 0.000001220940000000008086519912133522506399816617167150749792984412588073873640980634536939227676159288583
trimMaxSell = apply subscript rule on the number: 0.00000122094

minBuy: 0.0000011876018281898164606504723305566401008793488147663452991764643229544162750244140625
trimMinBuy = apply subscript rule on the number: 0.00000118760

After applying the new logic: 
IF trimMaxSell - trimMinBuy > 1 -> 0.00000122094 - 0.00000118760 > 1 -> 0.00000003334 > 1 -> FALSE
- False means display 6 decimals after the first non zero value
```

GrandSchtroumpf:
fix #1273

When order range delta is < 1 we truncate number after 6 decimals.

## Test cases :
- Create a disposable range strategy with theses prices
- Hover graph in strategy card to see prices (price is after sdk approximation)

**Both below 0**:
- Form: min 0.000000001 & max 0.01
- Card: min 0.0₈99999 & max 0.009999

**Above & below 0**:
- Form: min 0.99999 & max 1.00001
- Card: min 0.999989 & max 1.000009

**Both above 0**:
- Form: min 10.0005 & max 11.0000001
- Card: min 10.000499 & max 11.00


ashachaf (approved):

zavelevsky (approved):

e2e/screenshots/strategy/overlapping/Overlapping/editPrices/form.png
tiagofilipenunes, line 1: Just flagging that seems like the screenshot was taken before user balance query was finished. The BNT deposit error shouldn't be there and the deposit warning doesn't show up because of it

tiagofilipenunes (approved):

PR #1366, opened 2024-07-18
[Activity] Prepare frontend for backend update

### LINKED ISSUE #1365: [Activity] Link to pagination to backend
Instead of doing pagination on the frontend we want to use the backend.

### Search Params
The backend should support these new search params: 
- `start`: Filter out activities that happens before this unix timestamp
- `end`:  Filter out activities that happens after this unix timestamp
- `action`: List of actions to filter
- `pairs`: List of pairs to filter (this might replace token0, token1 in the future)
- `limit`: amount of activity to send to the client
- `offset`: first index of the activity

It would look like that : 
```typescript
interface BackendSearchParams {
  actions: string[];
  strategyIds: string[];
  token0: string;
  token1: string;
  pairs: string;  // `${base}_${quote},${base}_${quote}`
  start: number;
  end: number;
  limit: number;
  offset: number
}
```

### Meta params
Endpoint: `/v1/activity-meta`

In addition we'll need meta information about the query (without limit&offset)
- `size`: total amount of activity for this query
- `actions`: list of actions available in this query
- `pairs`: list of pairs available in this query
- `strategies`: list of strategy ids with their pairs to 

```typescript
export interface BackendActivityMeta {
  size: number;
  actions: ActivityAction[];
  strategies: Record<string, [string, string]>;
  pairs: [string, string][];
}
```

Note: we need to match the id with the pair because of the design of the filter for ids : 
image

### Forward the search params to the backend:
Here is what we need to do to forward params from frontend to backend :

| Frontend Params | Backend Params | Query                          |
|-----------------|----------------|--------------------------------|
| ids             | strategyIds    | strategyIds = ids              |
| actions         | actions        | actions = action               |
| pairs           | token0         | token0 = pairs[0]              |
| pairs           | token1         | token1 = pairs[1]              |
| start           | start          | start = Unix(start) + UTCDelta |
| end             | end            | end = Unix(end) + UTCDelta     |
| limit           | limit          | limit = limit \|\| 10          |
| offset          | offset         | offset = offset \|\| 0         |

### Remove local filtering
Do not use `ListProvider` for Activity nor Pagination component.
Forward the size of the query to the Pagination component.

GrandSchtroumpf:
fix #1365

Use backend for pagination & filter : 
- [x] actions
- [x] ids
- [x] pairs
- [x] owner
- [x] start/end

DO NOT MERGE before backend is deployed on ethereum


zavelevsky (approved):
Code LGTM

ashachaf:
- [ ] general notifications are off the charts
https://github.com/user-attachments/assets/ded3eb5d-1b05-4862-b313-73f892a024df

ashachaf:
- [ ] when applying filter the response fails
https://github.com/user-attachments/assets/00b56f96-e504-4b78-a32e-a4681ad3bd39

- [ ] removing filter cause the response to fail
https://github.com/user-attachments/assets/671be336-baf6-45f4-9695-8973217b8c6d

PR #1373, opened 2024-07-24
[E2E] set localstorage instead of using debug page

GrandSchtroumpf:
Reduce each test by 5-10sec


e2e/utils/DebugDriver.ts
```diff
@@ -22,6 +22,32 @@ export const setupFork = async (testInfo: TestInfo) => {
...
+    localStorage.setItem('carbon-v1.1-isUncheckedSigner', 'true');
+    localStorage.setItem(
+      'carbon-v1.1-carbonControllerAddress',
+      '"0xC537e898CD774e2dCBa3B14Ea6f34C93d5eA45e1"'
+    );
+    localStorage.setItem(
+      'carbon-v1.1-voucherContractAddress',
+      '"0x3660F04B79751e31128f6378eAC70807e38f554E"'
+    );
+  }, rpcUrl);
```
tiagofilipenunes, line 48: I would prefer if these values come from a separate initialLocalStorage file to make it easier to update with more initial localStorage data

e2e/utils/DebugDriver.ts
```diff
@@ -22,6 +22,32 @@ export const setupFork = async (testInfo: TestInfo) => {
   process.env[`TENDERLY_FORK_ID_TEST_${testInfo.testId}`] = fork.id;
 };
 
+export const setupLocalStorage = async (page: Page, testInfo: TestInfo) => {
+  const forkId = process.env[`TENDERLY_FORK_ID_TEST_${testInfo.testId}`];
+  if (!forkId)
+    throw new Error('Fork should be created before setting theRC URL');
```
tiagofilipenunes, line 28:
```suggestion
    throw new Error('Fork should be created before setting the RPC URL');
```

e2e/utils/DebugDriver.ts
```diff
@@ -22,6 +22,32 @@ export const setupFork = async (testInfo: TestInfo) => {
...
+  // We need to be on a page to set localstorage so we create an empty page
+  await page.route(testInfo.testId, (route) => {
+    return route.fulfill({ status: 200, contentType: 'text/plain', body: '' });
+  });
+  await page.goto(testInfo.testId);
+
+  return page.evaluate((rpc) => {
+    // each value is stringified to match lsservice
+    localStorage.setItem('carbon-v1.1-tenderlyRpc', `"${rpc}"`);
+    localStorage.setItem('carbon-v1.1-isUncheckedSigner', 'true');
```
tiagofilipenunes, line 39: Why not set imposterAccount here too and remove the need for debug.setImposter? Would probably save some more time and avoid visiting the debug page altogether for some tests

tiagofilipenunes (changes requested):
Nice 👍 - good improvement and more initial data for the local storage can be added over time to cut down the test duration further :) Commented some suggestions

e2e/utils/DebugDriver.ts
GrandSchtroumpf: We still need to get funds. The tenderly RPC was the most critical because it requires a reload. We can try to get rid of the need for debug page in another PR.

tiagofilipenunes: Oh, you're right. Agree with doing it in another PR

e2e/utils/DebugDriver.ts
```diff
@@ -25,27 +26,20 @@ export const setupFork = async (testInfo: TestInfo) => {
...
-      '"0xC537e898CD774e2dCBa3B14Ea6f34C93d5eA45e1"'
-    );
-    localStorage.setItem(
-      'carbon-v1.1-voucherContractAddress',
-      '"0x3660F04B79751e31128f6378eAC70807e38f554E"'
-    );
-  }, rpcUrl);
+    for (const [key, value] of Object.entries(storage)) {
+      localStorage.setItem(`carbon-v1.1-${key}`, JSON.stringify(value));
+    }
```
tiagofilipenunes, line 41: The key formatter is prone to change, could you use the lsService here instead?

tiagofilipenunes: Looks good now

tiagofilipenunes (approved):

PR #1374, opened 2024-07-24
Display fiat on budget description

### LINKED ISSUE #1244: Overlapping - adding budget distribution display currency value
currently, when a user is creating/editing/depositing/withdrawing in/from an overlapping strategy, the app shows the expected token amount to use for the 2nd token/budget.
image

the indication does not include the display currency value of the token amount.

please add the display currency value of the tokens in this way:
current: 
You will deposit 10.00 USDC from your wallet to the strategy.
You will deposit 0.002622 ETH from your wallet to the strategy.
updated:
You will deposit 10.00 USDC ($10.00) from your wallet to the strategy.
You will deposit 0.002622 ETH ($9.89) from your wallet to the strategy.
image

GrandSchtroumpf:
fix #1244 

image

Test case : 
- Create an overlapping strategy with one token with fiat and one without
- Enter a budget and see description: one has fiat price, the other not


ashachaf (approved):

tiagofilipenunes (approved):
LGTM

PR #1380, opened 2024-07-30
[Activity] Increase refresh interval & limit

### LINKED ISSUE #1379: General notification - increase interval and amount
Please make the following changes to the general notifications:
- increase pull interval from 30 seconds to 5 minutes
- increase notifications to display from 8 to 25
reason for this update is due to the fact that the api fetch new data on a 5min interval which means with the current implementation we would show 8 notifications in a 30second timeframe and then have 4:30 minutes with nothing.

GrandSchtroumpf:
fix #1379


ashachaf:
@GrandSchtroumpf you can increase the notification amount to 50.
50 notifications in 5 minutes = notification will be shown every 6 seconds (on average)

tiagofilipenunes (approved):
LGTM 👍

Note: If the API is offline, the query continuously tries to fetch with new keys (every ~7 seconds when running it locally on mine)

PR #1382, opened 2024-07-31
Hide chain selector if networks length is one

tiagofilipenunes:
Check networks.length is less than two and if so, hide the chain switch selector.


src/components/core/menu/mainMenu/MainMenuRightChainSelector.tsx
```diff
@@ -17,7 +17,7 @@ interface Props {
 
 export const MainMenuRightChainSelector: FC<Props> = ({ networks }) => {
   const activeNetwork = networks.find((network) => network.isCurrentNetwork);
-  if (!activeNetwork) return;
+  if (!activeNetwork || networks.length === 1) return;
```
zavelevsky, line 20: safer to check networks.length < 2

tiagofilipenunes: activeNetwork would be false if network.length === 0, but I'll change. Either works for me

zavelevsky (approved):

PR #1383, opened 2024-07-31
add activity documentation

GrandSchtroumpf:


docs/activity.md
```diff
@@ -0,0 +1,87 @@
...
+```typescript
+const { pairs } = await fetch('https://api.carbondefi.xyz/v1/activity/meta?actions=delete').then(res => res.json());
+```
+
+
+Get the last 10 activities from user 0x123:
+```typescript
+const { size } = await fetch('https://api.carbondefi.xyz/v1/activity/meta?owner=0x123').then(res => res.json());
+const offset = size - 10;
+const activities = await fetch(`https://api.carbondefi.xyz/v1/activity/meta?owner=0x123&limit=10&offset=${offset}`);
```
zavelevsky, line 80: no /meta

docs/activity.md
```diff
@@ -0,0 +1,87 @@
...
+  end?: number;
+}
+```
+
+
+## Activity Meta
+The meta endpoint uses the same params as the activity list without the pagination params (limit & offset)
+
+### Example
+Get all strategy IDs owned by user 0x123:
```
zavelevsky, line 63: these are all the strategies ever owned by the user, correct? Not just currently

docs/activity.md
```diff
@@ -0,0 +1,87 @@
...
+```typescript
+const { strategies } = await fetch('https://api.carbondefi.xyz/v1/activity/meta?owner=0x123').then(res => res.json());
+/** strategies is a Record of ID -> Pair */
+const ids = Object.key(strategies);
+```
+
+Get all the pairs of strategies that have been delete:
```
zavelevsky, line 70: deleted

zavelevsky (approved):

PR #1392, opened 2024-08-01
[Input] Use props id instead of inputId

GrandSchtroumpf:
Use props id to compare activeElement


tiagofilipenunes (approved):
Code LGTM 👍

Tried different input in limit, recurring, overlapping to test InputLimit, InputRange and InputBudget, and found no issues.

PR #1403, opened 2024-08-07
Fix pagination next page issues

### LINKED ISSUE #1402: [BUG] Cannot go past the 3rd page activities
The activity limit and offset searchParams used for activity are parsed to strings, then used to strings and not converted to numbers. This leads to a behaviour in the pagination where moving to the next page does the following:

Current: 1->2->102->10102...
Expected: 1->2->3->4...

tiagofilipenunes:
Fix #1402


GrandSchtroumpf (approved):
Approving for the sake of fixing production. Could you find out why we don't get a number in the first place in another PR?

PR #1406, opened 2024-08-08
Fix limit offset parse

tiagofilipenunes:
Currently all numbers in searchParams are being parsed to string and verified as a string number. 

This PR adds a special case for the limit and offset search params which are only used for pagination and do not require to be converted to string.


GrandSchtroumpf (approved):

GrandSchtroumpf:
It's ok, but I would rather have something more generic to manage numbers in the URL

GrandSchtroumpf (approved):

PR #1407, opened 2024-08-08
fix filter & empty state

### LINKED ISSUE #1395: [BUG] Activity - update the "no result" view
Flow:
1. go to the activity tab
2. apply filter which result in 0 rows
expected: https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=14577-41518&t=nEEg2lqsm0pAfQSS-4
actual: 
filters are gone and user cannot reset them
image

GrandSchtroumpf:
fix #1395

- Pair filter is now working
- Empty state leave the filters

Test case: "Pair filter"
- go to explorer with wallet address: 0x5f7a009664b771e889751f4fd721adc439033ecd
- go to activity tab
- select pair DAI/USDC
- filter works
- add USDC/BBS
-> It only display USDC/BBS, while it should display both. This is because backend response is not correct.

Test case "Empty State"
- go to explorer with wallet address: 0x5f7a009664b771e889751f4fd721adc439033ecd*
- filter by date from 1st of July to 2nd of July
- Display empty state, but filter are accessible

### DIFF
--- a/e2e/utils/modal.ts
+++ b/e2e/utils/modal.ts
@@ -32,6 +32,7 @@ export const checkApproval = async (page: Page, tokens: string[]) => {
 
 export const waitTooltipsClose = async (page: Page) => {
   const selector = '[data-testid="tippy-tooltip"]';
+  await page.mouse.move(0, 0);
   const tooltips = await page.locator(selector).all();
   for (const tooltip of tooltips) {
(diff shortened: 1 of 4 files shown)


tiagofilipenunes (approved):
LGTM

PR #1410, opened 2024-08-08
Improve display on approval error message

### LINKED ISSUE #1346: [Approval] Approval error output is not correctly shown to user
When approving, if there is an error on the wallet or rpc side, this error is printed next to "Waiting for confirmation". 

As this error is hard to reproduce since it requires the approval to fail on either the rpc or the wallet side, one way to do it is by running the app locally and using the tanstack query debug tool:
1. Run app locally on any network
2. Go to trade page and use an ERC-20 to get the approval modal
3. Press "approve"
4. In react-query debug tools, trigger error state for the query which has a queryKey starting with "["chain", "approval"...]"
5. "{}" is output to the modal, next to "Waiting for Confirmation". In case of an rpc or walletError, you'd see the error message info in json where the "{}" is

image

tiagofilipenunes:
fix #1346 

- Approval message was showing a JSON.stringify of the error. 
- Since it comes from a tanstack-react query, the error.message can be displayed instead of the entire error object.
- Adjusted the layout of the approval modal so that the height expands as needed to accommodate the error and the error occupies a minimum width.

### DIFF
--- a/src/components/common/approval/ApproveToken.tsx
+++ b/src/components/common/approval/ApproveToken.tsx
@@ -21,7 +21,7 @@ import config from 'config';
 type Props = {
   data?: ApprovalTokenResult;
   isPending: boolean;
-  error: unknown;
+  error: Error | null;
   eventData?: StrategyEventOrTradeEvent & TokenApprovalType;
   context?: 'depositStrategyFunds' | 'createStrategy' | 'trade';
 };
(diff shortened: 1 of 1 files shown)


src/components/common/approval/ApproveToken.tsx
```diff
@@ -221,7 +221,9 @@ export const ApproveToken: FC<Props> = ({
           </span>
         )}
 
-        {error ? <pre>{JSON.stringify(error, null, 2)}</pre> : null}
+        {error ? (
+          <pre className="max-w-min text-wrap">{error.message}</pre>
```
GrandSchtroumpf, line 225: Maybe use `output` if it's based on an form result.

GrandSchtroumpf (approved):

src/components/common/approval/ApproveToken.tsx
tiagofilipenunes: Agree

PR #1426, opened 2024-08-14
[My Strategies] Search to only apply in overview tab

### LINKED ISSUE #826: Wrong message when filtering strategies and there's no results
### Overview

On the "Explorer" page, we display the same 'We couldn't find any strategies' message in two scenarios: when no strategies are found while a user searches for a specific token pair and wallet, and when there are strategies to be displayed but the user applied a filter that filtered them out, resulting in 0 results. In the latter case, to prevent confusion and the user from thinking that no strategies were found at all, we should encourage users to adjust their filter criteria.

On the "My Strategies" page, when the filters or a search input result in 0 results, the system displays "Strategy Not Found" without suggesting the user to alter their search input or adjust their filters.

### Explorer page

**Flow:**
- **IF** there are 0 search results 
-- Display  "We couldn't find any strategies" message (same as today) regardless of what filters are applied
- **ELSE IF** there are >0 search results but the user applied a filter that filtered them out, resulting in 0 results.
-- Display the following message [mockup](https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=9419%3A50772&mode=design&t=QymZpL3Tb7nRXmiN-1)

### My strategies page
**Flow:**
- **IF** user has 0 strategies
-- Display current "Create Your First Strategy" message (same as today)
- **ELSE IF** there are >0 strategies but the user applied a filter that filtered them out, resulting in 0 results.
--Display the following message [mockup](https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=9440-37357&mode=design&t=tPodZ4SRrjNrurqQ-0)

### LINKED ISSUE #1150: [Explorer] Filter&Sort shows in Portfolio tab but has no effect
Actual: Filter&Sort button shows in the Portfolio tab in the Explorer. Any selection doesn't apply to the Portfolio page.

Expect: No Filter&Sort button is shown, similarly to what is seen in the portfolio tab of the Strategies page.

### LINKED ISSUE #1425: Overview tab search functionality
### Overview
When a user searches for a strategy on the Overview tab within the My Strategies page, those search parameters should not carry over to the Portfolio tab.

**Steps to Reproduce:**

1. Connect a wallet that contains several strategies.
2. Navigate to the Overview tab on My Strategies page.
3. Enter search parameters in the search field that result in "no strategies found."
4. Switch to the Portfolio tab.

**Expected Result:**
The Portfolio tab should display the token distribution for the connected wallet.

**Actual Result:**
A "Create your first strategy" CTA is displayed instead.

tiagofilipenunes:
- Search to only apply in overview tab

fix #1425 
fix #1150 
fix #826


src/hooks/useStrategies.tsx
```diff
@@ -20,14 +21,40 @@ import { getCompareFunctionBySortType } from 'components/strategies/overview/uti
...
+
+  const setSearch = useCallback(
+    (value: string) => {
+      navigate({
+        search: (currentSearch: MyStrategiesSearch) => {
+          if (value) return { ...currentSearch, search: value };
+          const searchCopy = structuredClone(currentSearch);
+          delete searchCopy.search;
+          return searchCopy;
+        },
```
GrandSchtroumpf, line 49: I think you can do that :
```suggestion
        search: (currentSearch: MyStrategiesSearch) => {
          return { ...currentSearch, search: value ?? '' };
        },
```

src/hooks/useStrategies.tsx
```diff
@@ -100,6 +128,7 @@ export const useStrategiesWithFiat = (
 type StrategyCtx = ReturnType<typeof useStrategyFilter>;
 export const StrategyContext = createContext<StrategyCtx>({
   strategies: [],
+  filteredStrategies: [],
```
GrandSchtroumpf, line 131: I would rather rename to `allStrategies` & `strategies`

src/pages/explorer/type/overview/index.tsx
```diff
@@ -3,21 +3,30 @@ import { useStrategyCtx } from 'hooks/useStrategies';
...
+    />
+  ) : (
     <NotFound
       variant="error"
       title="We couldn't find any strategies"
-      text="Try entering a different wallet address or choose a different token pair or reset your filters."
+      text="Try entering a different wallet address or choose a different token pair."
       bordered
     />
   );
```
GrandSchtroumpf, line 24: I would do something like that instead: 
```jsx
const emptyProps = strategies.length
  ? { title: 'There are no results for your filter', text: '...' }
  : { title: 'We couldn't find any strategies', text: '...' };
const empty = <NotFound variant="error" bordered {...emptyProps} />
```

GrandSchtroumpf (commented):
I would rather re

src/hooks/useStrategies.tsx
tiagofilipenunes: I had it initially as this but I think calling `filteredStrategies` and `strategies` vs `strategies` and `allStrategies` respectively makes more sense as it not obvious when using the hook and reading the code that `strategies` is in fact filtered when allStrategies is not deconstructed from the hook's outputs:

```
const { strategies } = useStrategiesCtx() // NOT obvious that the strategies are filtered here
const { allStrategies } = useStrategiesCtx() // obvious that the strategies are NOT filtered here
```

vs


```
const { filteredStrategies } = useStrategiesCtx() // obvious that the strategies are filtered here
const { strategies } = useStrategiesCtx() // can be safely assumed that the strategies are NOT filtered here
```


Can change if you feel strongly about the first option

src/pages/explorer/type/overview/index.tsx
tiagofilipenunes: Agree

src/hooks/useStrategies.tsx
tiagofilipenunes: Agree

tiagofilipenunes: Used `Route.postSearchFilters` to remove the search param clearing logic from the navigate function

Costa769 (approved):

PR #1456, opened 2024-08-30
[Trade] History Chart

### LINKED ISSUE #1187: [Interactive Chart] - Trade Page
# Overview

With the expansion of token offering of exotic pairs on multiple chains, the existing charting library cannot provide a reliable solution to display price history to the users.
In addition, the current library limitation does not allow the charts to be an integrated and interactive part of the user strategy creation and management flows.

As a result, the goal of this task is to replace the current TradingView implementation with a proprietary charting solution across the entire app. 

# Designs

[All chart pages](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-65332&t=zfGNlQDrpwiFOze0-1)

# Implementation logic

- Replace tradingView chart with the proprietary interactive charts on every page/flow they are used
-- The interactive chart on each page will use the same date picker as the one on the simulate page.
- **Important:** 
To allow flexibility for our licensees, we are required to support both chart types, enabling each licensee to choose between integrating the new interactive charts or continuing to use TradingView ones. 
Meaning, the licensee will have the option to choose the current implementation of TradingView or to display the new implementation outlined in this issue. 

Support for both chart types should be configured individually for each chain as well. For example, interactive charts may be implemented for the Ethereum chain, while other chains could continue using TradingView charts.

For further details on managing interactive charts across different pages and strategies, please refer to the section below.

## Strategies: Limit/Range and Recurring:

The above strategies on the trade page and edit price pages as well as the deposit and withdraw pages will feature the new interactive chart with drag-and-drop functionality, as seen on the "Simulate" page, following these rules:

### Strategy creation
During the process of creating a new strategy
#### Charts display & interactivity rules
- **IF** we have price history
-- Show the interactive chart with the price indicators [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-68025&t=rVk3QVW3nAIhYNtb-1)
-- The chart should allow drag and drop functionality similar to the one  available on the simulate page

- **ELSE**, 
-- Collapse the interactive price chart upon page load.
-- Interactive chart without price history expended [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-73930&t=rVk3QVW3nAIhYNtb-1)

#### Populating price fields

We want to ensure that users understand how to interact with the interactive chart as soon as the trade page with limit/range and recurring strategies  loads, therefore:

- Upon loading the trade page with one of the above strategies, the price fields for limit, range and recurring strategies should be automatically filled using the default price input rules identical to those applied for recurring strategies on the simulation page.
```
Expected behavior (example)
-- Go to trade page
-- Select Limit/Range strategy option
-- Buy Low price field should be pre-populated with a value 
(calculate in the same way as we do in the simulate page, pre-populated value = market price *99%)
```

**Once a value is available, the interactive chart will display it as well and allow the drag and drop functionality already available on the simulate page**

## Strategies: Overlapping

Unlike the simulate where it only supports tokens with price history, on the strategy creation/edit/etc flows, we have different options based on the available data at hand. Additionally, apart from replacing the TradingView chart with the interactive chart, we plan to keep the range chart and move it  under the range tab on the interactive chart. Both charts behavior is described below

### Strategy creation
During the create strategy process

- **IF** we have market price and price history
-- Display price range chart by default on page load [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-68434&t=rVk3QVW3nAIhYNtb-1)
-- Show the interactive chart with the price points [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-66202&t=rVk3QVW3nAIhYNtb-1)
-- The interactive chart should allow drag and drop functionality similar to the one available on the simulate page

- **ELSE, IF** we have no market price and no price history
-- Prompt the user to set a market price on page load (Setting the market price uses the same logic as currently implemented)
-- Display price range chart by default on page load [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-73930&t=rVk3QVW3nAIhYNtb-1)
-- Show interactive chart without price history which remain unchanged in case the user updates the market price. 
 [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-73930&t=rVk3QVW3nAIhYNtb-1)

- **ELSE, IF** we have market price but no price history
-- Display price range chart by default on page load
-- The interactive chart maintain the same behavior as described in the section above

- **ELSE, IF** we have no market price but have price history
-- Same logic as in section with no market price and no history but the interactive chart will display the price indicators and the new market price that was set by the user.[mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-67589&t=rVk3QVW3nAIhYNtb-1)

### Editing existing market price
The edit market price button on both strategy creation and edit prices pages of the overlapping strategies has been relocated to the interactive chart section. The process for editing existing market prices will remain largely unchanged. The edit price form will still be used but will now be displayed in the center of the page. [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-66639&t=VbAMxOSHJm5SlQzL-1)
After the market price update, the following logic will be applied:

- **IF** there is a price history:
-- The user remains on the same range chart or interactive chart tab they were on when they clicked the edit market price button.
-- The market price is updated on both range and interactive charts.[Interactive chart post update mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-67589&t=VbAMxOSHJm5SlQzL-1)

- **ELSE:**
-- Market price is updated on the range chart, and the focus shifts to the range chart tab.
-- The market price is not updated on the interactive chart, and the interactive chart retains its no-price-history design.

## Strategies: Spot

Since the chart when the spot strategy is selected  on trade page lacks price points and spreads, users can only modify the date range of the interactive chart. Regarding the display rules, the following logic should be applied:

- **IF** there is a price history
-- Show history tab with interactive chart on page load

- **ESLE**
-- Show history  chart without the price history

## Mobile implementation

Interactive charts will be introduced on mobile as well. However, due to the challenge of interacting with price points on mobile devices, the following logic will be applied:

For all pages the interactive chart is displayed on:
- The interactive charts will show the same data we show on web (price points, market price, etc) [Overlapping mockup 1](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-75326&t=rVk3QVW3nAIhYNtb-1), [Overlapping mockup 2](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-75687&t=rVk3QVW3nAIhYNtb-1), [Limit mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-76083&t=rVk3QVW3nAIhYNtb-1)
- Direct editing of prices will not be allowed on the interactive chart.
- On pages with range charts and pages where interactive charts display price points, a notification will be displayed below the chart, explaining that the drag and drop feature is available only on web.
-- If there are other notifications displayed below the  chart (such as on deposit or withdraw pages), they should be hidden, and only the notification above should be shown.

### LINKED ISSUE #1452: [Interactive Chart] - Edit Prices Page
## Implement interactive chart on edit prices page

The following outlines the interactive chart functionality on edit prices page for different strategies.

**Limit/Range and Recurring Strategies**

On the edit prices page, the interactive chart will display the current strategy price points, and represent the existing strategy. The drag-and-drop functionality of this interactive chart is the same as described in the limit/range and recurring strategy creation issue. [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-71677&t=rVk3QVW3nAIhYNtb-1)

**Concentrated Strategies**

Similar to the edit prices on limits, ranges, and recurring strategy pages, the interactive chart will display only the current strategy price points, excluding any previously modified price points. The both charts (range and price history) functionality is the same as described in the overlapping strategy creation issue. [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-71254&t=UBGR4dcBDNn0kPPU-1)

### LINKED ISSUE #1453: [Interactive Chart] - Deposit&Withdraw Pages
## Implement interactive chart on deposit&withdraw pages

The following outlines the interactive chart functionality on the deposit and withdrawal pages for different strategies.

**Limit/Range and Recurring Strategies**

The logic for the interactive chart on the deposit and withdraw pages is similar to that of the set prices page, but it does not include the drag-and-drop functionality. 

- **IF** we have price history
-- Show the interactive chart with the price points
-- Drag and drop functionality should be disabled
-- Show the text below the chart that will indicate to the user that the drag and drop functionality is available on the edit prices page [deposit mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-70782&t=rVk3QVW3nAIhYNtb-1), [withdraw mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-69849&t=rVk3QVW3nAIhYNtb-1)
--- A warning below the chart will state that the price points are relevant to the current strategy and do not include any previous prices if they have been changed. (Text TBD)
- **ELSE**, 
-- Collapse the interactive price chart upon page load.

**Concentrated Strategy**

During the process of depositing or withdrawing funds to/from an existing concentrated strategy

- **IF** we have both a market price and price history
-- Display Range chart tab by default
--- Drag and drop functionality on the range chart should be disabled
--- Show the text below the range chart that will indicate to the user that the drag and drop functionality is available on the edit prices page(same text that is displayed under the interactive chart)
-- The interactive chart behavior is similar to the deposit &withdraw pages on limit/range/recurring strategies described in the above section [deposit mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-70310&t=rVk3QVW3nAIhYNtb-1), [withdraw mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-69366&t=rVk3QVW3nAIhYNtb-1)

- **ELSE, IF** we have market price but no price history
-- Range chart behavior is similar to the section above
-- Show interactive chart without price history

- **ELSE, IF** we have no market price but do have a price history
-- Display interactive chart tab by default
-- Interactive chart behavior is similar to the section where we have both a market price and price history
-- Range chart without market price design(similar to the following [mockup](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=15862-74891&t=rVk3QVW3nAIhYNtb-1) but with no "Please provide a price" text in the notification 

- **ELSE**, IF** we have no market price and no price history 
-- Collapse the range and interactive price charts upon page load.

GrandSchtroumpf:
fix #1187
fix #1453
fix #1452


tiagofilipenunes (changes requested):
Amazing improvement to the app! 

## Review Part 1

### Create strategy:
- [ ] Can src/components/simulator/input/d3Chart/recurring/D3ChartDisposable.tsx move to src/components/simulator/input/d3Chart/disposable/D3ChartDisposable.tsx
- [ ] Limit/Range and Recurring: Chart handle allows to set negative numbers
- [ ] Limit/Range: In range option, moving the range up or down and clicking "Use Market" so that the range becomes inverted doesn't invert the numbers in the form
- [ ] Concentrated: Max spread/fee tier is not adjusting when moving the ranges
- [ ] Spot: Can't input a number with decimals in the source or target input, try "2.2", "0.2" for example

- [ ] Consider moving the chart functionality away from the components/simulator folder as it is now used elsewhere in the app

src/components/simulator/input/d3Chart/disposable/D3ChartDisposable.tsx
```diff
@@ -0,0 +1,143 @@
...
+      prices.buy.min,
+      prices.sell.max,
+      prices.sell.min,
+      yScale,
+    ]
+  );
+
+  const labels = {
+    buy: {
+      min: prettifyNumber(prices.buy.min, { decimals: 4 }),
```
tiagofilipenunes, line 94: prettifyNumber(value, { decimals: 4}) happens a lot throughout the app. Consider creating a utils function to keep the label number formatting consistent

src/components/strategies/common/StrategyChartHistory.tsx
```diff
@@ -0,0 +1,187 @@
...
+  if (isError) {
+    return (
+      <NotFound
+        variant="error"
+        title="Well, this doesn't happen often..."
+        text="Unfortunately, price history for this pair is not available."
+      />
+    );
+  }
+  if (!data?.length) return null;
```
tiagofilipenunes, line 162: Could you check this condition with isError and show the NotFound component for it?

tiagofilipenunes (changes requested):
### Review part 2
- Edit price and edit budget chart expands to width larger than viewport
image
image

- The strategyBlock component in the individual strategy page expands beyond the desired height when there is no data for the chart
image

src/components/simulator/input/d3Chart/disposable/D3ChartDisposable.tsx
GrandSchtroumpf: I don't think this is a good idea to create a wrapper of `prettifyNumber` just for one param in the option. I find it already difficult to work with all the number function, I think it'll just make it harder.

tiagofilipenunes (changes requested):
- [ ] Chart handle can go to negative numbers and eventually is hidden from the user

https://github.com/user-attachments/assets/ca458601-fd9c-4dd0-8cec-057dc62f17fe

PR #1457, opened 2024-08-30
Add config override to debug page

### LINKED ISSUE #1397: Debug page should provide a way to use the app with other chains
We should expose a way to paste a config json and/or override specific entries to get the app to run with another chain.
This is needed for cases where carbon contracts are to be deployed to a new chain but not the app - and we'd want to create a few test strategies for the benefit of bot research

tiagofilipenunes:
- Add config override to /debug page
- Allow to load default config, save a config or reset
- Allow partial override of the config
- Show error message if config JSON cannot be parsed

fix #1397


src/components/debug/DebugConfig/DebugConfig.tsx
```diff
@@ -0,0 +1,92 @@
...
+      {!!error && (
+        <Warning isError message={error} htmlFor="custom-config-json" />
+      )}
+      <Button data-testid="save-config" type="submit" fullWidth>
+        Save
+      </Button>
+      <Button type="button" onClick={handleLoadDefault} fullWidth>
+        Load Default Config
+      </Button>
+      <Button type="button" onClick={() => saveConfigOverride()} fullWidth>
```
GrandSchtroumpf, line 87: you can use the `type="reset"` if you want and listen on the onReset event on the form.

src/components/debug/DebugConfig/growwrap.module.css
```diff
@@ -0,0 +1,30 @@
+.growwrap {
```
GrandSchtroumpf, line 1: I think you can use the `rows` attribute on `textarea` instead

src/components/debug/DebugConfig/DebugConfig.tsx
tiagofilipenunes: Done

GrandSchtroumpf (approved):

PR #1459, opened 2024-09-02
Fix distribution budget on edit

GrandSchtroumpf:
Remove unnecessary fields on edit


tiagofilipenunes (approved):
LGTM.

Not setting budget will not impact budget distribution when updating strategy by the SDK:

https://github.com/bancorprotocol/carbon-sdk/blob/main/src/strategy-management/Toolkit.ts#L1016

tiagofilipenunes (approved):
Test checks that correct inputs are sent to carbonSDK.updateStrategy, which matches the behaviour expected on the [sdk side](https://github.com/bancorprotocol/carbon-sdk/blob/70c8011ec7bf7364e7485767dcfb933121f8d10b/src/strategy-management/Toolkit.ts#L1000) with undefined sellBudget, sellPriceLow, sellPriceHigh and sellPriceMarginal. Doesn't check the output of the function in this case but the SDK tests cover it.

PR #1463, opened 2024-09-02
Fix approval

GrandSchtroumpf:
avoid precision error


ashachaf (approved):

tiagofilipenunes (approved):
Code LGTM

Node: `getDeposit` and `getWithdrawal` should also check that budget is non-zero so they return the same value (they’re only used for display)

PR #1468, opened 2024-09-04
Hide ROI by default

### LINKED ISSUE #1467: Overview card - remove ROI
For the SEI deployments only

Given the ROI information is currently not available for these deployments, please remove the entire information box from the overview card.

Please set a configuration value at this stage to control if we would like to hide/expose the roi information.
For Ethereum and Celo app, keep displaying it as we do today.
For Sei, hide it so it looks like the mockup below

Mockup:
https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=17328-64482&t=XImk0lYTPNFrNebR-4

tiagofilipenunes:
- Create optional config param
- If param is false or undefined, hide StrategyBlockRoi and adjust StrategyBlockBudget to the full column width
- Update readme with new flag

Note: ROI block is hidden by default

fix #1467


ashachaf:
- [ ] sort by ROI should also be hidden
- explore / my strategies filter option should not expose roi
- IF hidden, use sort by total budget ascending
image

- [ ] strategy page - apply same rule
image

ashachaf:
- [ ]  updated default sort is missing the checkmark indication in the menu
image

ashachaf (approved):

GrandSchtroumpf (approved):

GrandSchtroumpf (approved):

PR #1469, opened 2024-09-05
Token Hint Selection - Add duplicate symbol warning to token list selection

### LINKED ISSUE #1106: Token selection hint
### Overview 

Currently, when a user selects a token for trading (as well as in strategy creation and explorer), we do not provide any specific hints to explain which address the selected token corresponds to. This becomes an issue when there are multiple tokens sharing the same symbol, even if they have different logos.

Therefore, if there are several tokens sharing the same symbol, we aim to expose the addresses associated with each of those tokens.

### Design
[mockup](https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=13955%3A45701&mode=design&t=uUZC3Fxyb8lWHXlu-1)

### Implementation logic

The following logic should be implemented for both single and token pair selection lists on trade, explorer, simulator and strategy creation pages:

**IF** there are multiple tokens with the exact same symbol in the token selection list:
```
example 1, case TRUE:
token1 = abc
token2 = abc
example 2, case FALSE:
token1 = abc
token2 = abcd
```

- Display the first and last 6 digits of the token address beneath each token sharing the same symbol.
- IF multiple token pairs have at least one token with the same symbol, show  both token addresses of each pair [mockup](https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=13955%3A45701&mode=design&t=uUZC3Fxyb8lWHXlu-1).
- Upon hovering over the warning icon, display a tooltip with a token's address or, in the case of a token pair, the addresses of both tokens.[mockup](https://www.figma.com/file/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?type=design&node-id=13955%3A45794&mode=design&t=uUZC3Fxyb8lWHXlu-1)
-- Upon clicking the link icon next to an address in the tooltip, open a new tab with the token address on etherscan.

tiagofilipenunes:
- Add logic to check duplicate symbol in token list
- Add duplicate symbol warning with address and tooltip

fix #1106


ashachaf:
- [ ] align the tool tip external link icon
image

- [ ] (optional) display the warning on the 2nd token even if the 1st was selected
flow:
user search for xyz
list shows 2 options with the warning
user select 1 option
user search for xyz
current: list shows 1 option (because the 2nd was selected and removed from the list) with no warning that there are multiple tokens with this symbol
updated: list shows 1 option with a warning that there are multiple tokens with this symbol

ashachaf (approved):

src/libs/modals/modals/ModalTokenList/useModalTokenList.ts
```diff
@@ -75,6 +75,17 @@ export const useModalTokenList = ({ id, data }: Props) => {
...
 
+  const duplicateSymbols = useMemo(() => {
+    return Object.values(
+      tokens.reduce((r: { [symbol: string]: string[] }, v) => {
+        (r[v.symbol] ??= []).push(v.symbol);
+        return r;
+      }, {})
+    )
+      .filter((g) => g.length > 1)
+      .flat();
```
GrandSchtroumpf, line 86: This is quite complex. Consider using a for loop : 
```ts
const isDuplicate = {};
const duplicates = new Set();
for (const symbol of Object.values(tokens)) {
  if (isDuplicate[symbol]) duplicates.add(symbol);
  isDuplicate[symbol] = true;
}
return Array.from(duplicates);
```

tiagofilipenunes: That's right - this is much simpler. Done

src/libs/modals/modals/ModalTokenList/ModalTokenListDuplicateWarning.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const isNativeToken = token.address === NATIVE_TOKEN_ADDRESS;
+
+  const duplicatedTokenTooltipElement = (token: Token) => {
+    if (isNativeToken) return 'This is the native token of the chain.';
+
+    return (
+      <>
+        It appears there might be multiple tokens with the same symbol. Please
+        ensure you select the correct address.
+        <br /> {token.symbol}:<br />
```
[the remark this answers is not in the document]
tiagofilipenunes: Done and checked other <br /> in the app that weren't inside a <p>

GrandSchtroumpf (approved):

PR #1474, opened 2024-09-09
[CI] Add check that PR head commit does not start with [CI]

tiagofilipenunes:
If PR head commit starts with [CI], do not commit screenshots from the e2e tests


GrandSchtroumpf (approved):
Would be nice to have something like `['[CI]', ...]includes(...)`, but I'm not sure if it's possible.

PR #1486, opened 2024-09-13
Show warning when fiat price is unavailable

### LINKED ISSUE #1443: Trade - Add USD price indication for tokens with no price
Currently:
IF token has no USD price, we remove the USD indication as it cannot be calculated.
image

Updated:
If token has no USD price, we should display an indication that USD value is not available
image

tiagofilipenunes:
fix #1443


src/components/strategies/common/InputBudget.tsx
```diff
@@ -145,7 +150,9 @@ export const InputBudget: FC<Props> = (props) => {
         </div>
         <div className="text-12 font-weight-500 flex min-h-[16px] flex-wrap items-center justify-between gap-10 font-mono">
           <p className="flex items-center gap-5 break-all text-white/60">
-            {fiatValue.gt(0) && prettifyNumber(fiatValue, { currentCurrency })}
+            {showFiatValue && prettifyNumber(fiatValue, { currentCurrency })}
+            {!hasFiatValue() &&
+              `Current ${selectedFiatCurrency} value unavailable`}
```
GrandSchtroumpf, line 155: Is there a case when you can have both ? If so I would rather have a ternary
Is there a case when you can have none ? If so what should be displayed ?

In any case I would rather move the output text as a variable above. I like to reduce code logic inside the template.

GrandSchtroumpf (approved):

src/components/strategies/common/InputBudget.tsx
tiagofilipenunes: You're right, done, it's much clearer now

GrandSchtroumpf (approved):

PR #1498, opened 2024-09-19
Improve search params

### LINKED ISSUE #1475: Explore better way to parse search params
Currently the search params are mostly parsed using validateSearch from an individual route. This has lead to problems namely with limit and offset from the activity pages. 

Explore a more generic way to handle and parse numbers so that they don't come as strings or so that a case-by-case validation and parsing does not need to be defined.

GrandSchtroumpf:
fix #1475


src/pages/explorer/type/activity/index.tsx
```diff
@@ -4,16 +4,16 @@ import { useExplorerParams } from 'components/explorer';
 import { QueryActivityParams } from 'libs/queries/extApi/activity';
 
 export const ExplorerActivityPage = () => {
-  const { type, slug } = useExplorerParams();
+  const { type, slug } = useExplorerParams('/explore/$type/$slug/activity');
   const params: QueryActivityParams = {};
   if (type === 'wallet') params.ownerId = slug;
   if (type === 'token-pair') {
-    const [base, quote] = slug.split('_');
+    const [base, quote] = slug!.split('_');
```
tiagofilipenunes, line 11:
```suggestion
    const [base, quote] = slug.split('_');
```

src/libs/routing/utils.ts
```diff
@@ -113,15 +23,11 @@ export const validArrayOf = (schema: v.GenericSchema<string, any>) => {
   return v.array(schema);
 };
```
tiagofilipenunes, line 25: Passing the type improves inferring types with arrays, instead of just showing `any[]`

```
export const validArrayOf = <T>(schema: v.GenericSchema<string, T>) => {
  return v.array(schema);
};
```

src/components/explorer/suggestion/SuggestionList.tsx
```diff
@@ -29,7 +29,6 @@ export const SuggestionList: FC<Props> = (props) => {
             className="px-30 flex cursor-pointer items-center space-x-10 py-10  hover:bg-white/20 aria-selected:bg-white/10"
             to="/explore/$type/$slug"
             params={params}
-            search={{}}
```
tiagofilipenunes, line 32: Nice, the empty objects on `search` and `params` can now be removed from these files too:
- src/components/core/menu/mainMenu/MainMenuLeft.tsx
- src/components/core/menu/mobileMenu/MobileMenu.tsx
- src/components/strategies/StrategyPageTabs.tsx
- src/components/trade/TradeExplorerTabs.tsx

tiagofilipenunes (approved):
Code LGTM, great improvement with a better integration between Tanstack Router and valibot. Just some minor suggestions. Tested the app thoroughly and no issues found.

tiagofilipenunes (approved):

PR #1523, opened 2024-10-01
Fix simulator recurring handling default values

### LINKED ISSUE #1522: [Simulator] Bug when changing token selection with a token that has no market price in simulator recurring
1. Go to simulate page and choose recurring strategy
2. Set a token in the token selection that has no market price (example: BBS)

Current: Try changing tokens again. The change is reflected.
Expected: Changing the tokens again works.

tiagofilipenunes:
- Stop returning when market price does not exist
- If market price does not exist, set min and max to 0

fix #1522


src/pages/simulator/recurring/index.tsx
```diff
@@ -50,13 +50,17 @@ export const SimulatorInputRecurringPage = () => {
...
+        ? new SafeDecimal(marketPrice)
+            [operation](marketPrice * multiplierMax)
+            .toFixed()
+        : '';
+
+      const min = marketPrice
+        ? new SafeDecimal(marketPrice)
+            [operation](marketPrice * multiplierMin)
+            .toFixed()
+        : '';
```
GrandSchtroumpf, line 63: This is very difficult to read, I think it is worth refactoring a little.

tiagofilipenunes: Refactored

GrandSchtroumpf (approved):

PR #1540, opened 2024-10-08
Update icons and animations to brand colours

tiagofilipenunes:
- Update portfolio page icons
- Update checkbox to gradient style
- Update Connect Wallet loading
- Update Create Strategy button


Costa769 (approved):

.env.sample
```diff
@@ -10,11 +10,14 @@ VITE_SDK_VERBOSITY=
 # For usage with contracts version < 5, without the enhanced range for trade by source. In that case - set this to true
 VITE_LEGACY_TRADE_BY_SOURCE_RANGE=
 
+# Opt-in to the gradient branding by setting this flag to true
+VITE_USE_GRADIENT_BRANDING=
```
GrandSchtroumpf, line 14: I don't think .env variable is the good place to manage this. Else it'll become the place for every UI stuff. I would recommend using config.ui instead

src/components/common/Checkbox/Checkbox.tsx
```diff
@@ -9,16 +9,18 @@ type Props = JSX.IntrinsicElements['button'] & {
 
 export const Checkbox: FC<Props> = ({ isChecked, setIsChecked, ...attr }) => {
   const classNames =
-    'size-18 rounded-4 border-2 flex items-center justify-center cursor-pointer';
-  const borderColor = isChecked
-    ? 'border-primary bg-primary'
-    : 'border-white/60';
+    'size-18 rounded-4 flex items-center justify-center cursor-pointer';
+  const borderBgColor = isChecked
+    ? 'bg-gradient-to-r from-primaryGradient-first to-primaryGradient-middle'
```
GrandSchtroumpf, line 14: Would be worth creating a dedicated class for that.

src/index.css
```diff
@@ -69,8 +72,45 @@
...
+  input[type='checkbox']:not(:checked):hover {
+    border-color: rgb(255 255 255 / 0.8);
+  }
+  input[type='checkbox']:checked {
+    border: 0px;
+    color: rgba(22, 22, 22, var(--tw-text-opacity));
+    background-image: linear-gradient(
+      90deg,
+      var(--gradientFirst) 0%,
+      var(--gradientMiddle) 100%
```
GrandSchtroumpf, line 91: You don't want the gradientLast here ?

src/index.css
```diff
@@ -69,8 +72,45 @@
...
+    filter: brightness(1.2);
+  }
+  input[type='checkbox']::before {
+    display: none;
+    content: '';
+    position: absolute;
+    top: 60%;
+    left: 50%;
+    transform: translate(-50%, -50%);
+    width: 65%;
```
GrandSchtroumpf, line 104: You should consider using a grid of something, this is very manual positioning

src/index.css
```diff
@@ -69,8 +72,45 @@
...
+    aspect-ratio: 1 / 1;
+    pointer-events: none;
+    background-image: url('assets/icons/check.svg');
```
GrandSchtroumpf, line 107: It feels like you should create a checkbox component to manage this with a real svg icon instead of using images

tailwind.config.ts
```diff
@@ -1,6 +1,7 @@
 import type { Config } from 'tailwindcss';
 import { formatRgb } from 'culori';
 import plugin from 'tailwindcss/plugin';
+import 'dotenv/config'; // TODO: we can remove this with node@20.6.0
```
GrandSchtroumpf, line 4: Hmm are we not using node@20 already ?

src/components/common/Checkbox/Checkbox.tsx
tiagofilipenunes: Done

tailwind.config.ts
tiagofilipenunes: Copied this from playwright and left the comment as it felt out of scope to address removing dotenv. See this PR: https://github.com/bancorprotocol/carbon-app/pull/1543

src/index.css
tiagofilipenunes: Done

src/index.css
tiagofilipenunes: `gradientLast` is used for the hover effect in the buttons and using gradientFirst to gradientMiddle makes the checkbox background look just like the buttons' background

GrandSchtroumpf (approved):

src/index.css
tiagofilipenunes: Created an issue to make the checkbox implementation uniform throughout the app: https://github.com/bancorprotocol/carbon-app/issues/1544

PR #1574, opened 2024-11-12
Feature flag

GrandSchtroumpf:
User can (un)select the feature flag in the debug page. Flag is store in the local storage.

image

If there is no feature flag display a message.
image

What this issue doesn't do : 
- display a warning when user has old feature flag in the local storage
- display a notification to the user when we deprecate the feature flag
Should it be include in the PR ?


zavelevsky:
> * display a warning when user has old feature flag in the local storage
> * display a notification to the user when we deprecate the feature flag
>   Should it be include in the PR ?

It is safe to assume that we will utilize this feature for now only with users with whom we have close and direct communication. So it is ok to leave these out, as we will let these users know what's going on and it's best to not require any action on their end when the feature's availability changes.

zavelevsky (approved):
Looks great!

PR #1588, opened 2024-12-02
[Performance] Trade Pair filter

GrandSchtroumpf:
The main goal of this PR was to identify what could impact performance on the explorer page, and see how we can reduce it to its maximum with the same UX. It comes at a devX cost, therefore it shouldn't be used a reference for future component but as a toolbox of actionable items when performance is poor on a specific component.

Before: 415ms per keydown event
image

After: 7-30ms per keydown event
image

- [x] Removed Sentry Replay
- [x] Removed the useKeyPress that was always listening on all keypress event. Start listening only when modal is open
- [x] Use button instead of <Link /> element to reduce the amount of hooks used
- [x] Use virtualize scroll
- [x] Do not update state on keydown
- [x] Manage style state with CSS selectors when possible
- [x] Use `content-visibility: auto` to reduce Layout impact for items not visible
- [x] Memorize `PairLogoName`
- [x] `usePair` keep cache into the state instead of reading the LS on each rerendering

Breaking changes : 
- Now pressing Escape to close a Modal only work on desktop, not on mobile

What to test : 
- Go to explorer page
- Filter out some tokens
- Select one
- Refresh
- Clear input field
- Focus in/out of the input


tiagofilipenunes (changes requested):
- Remove Sentry Replay is a good idea, the replays can be useful but if the user uses an adblocker they won't be sent anyways
- That key press hook was a good catch and a very odd implementation of a key event listener...
- I'm okay with replacing Link in the suggestionList
- Virtualization looks really good
- Performance significantly improved, great work!


- [ ] Scrolling with the arrow past the end doesn't go all the way to the top for me, for long lists (try 'ETH')

PR #1596, opened 2024-12-11
Upgrade to React 19

### LINKED ISSUE #139: Look into framer-motion latest version bug
animations are stuttering when updating framer-motion to latest version

### LINKED ISSUE #1175: [Chore] Replace Graphemer by Intl.Segmenter()
Graphemer is used in the app to perform non-trivial split of strings, for strings such as >1-char emojis used in ENS addresses.

[Intl.Segmenter](https://caniuse.com/?search=Intl.Segmenter) is now widely supported since Firefox 125 (16th April) has included it. Graphemer can be easily replaced by Intl.Segmenter. 

Before:

```ts
  const splitter = new Graphemer();
  const stringArray = splitter.splitGraphemes(stringToShorten);
```

After:

```ts
  const segmenter = new Intl.Segmenter();

  const stringArray = [...segmenter.segment(stringToShorten)].map(
    (segmentData) => segmentData.segment
  );
```

GrandSchtroumpf:
fix #139
fix #1175

- [x] Remove unnecessary libraries heavy on the bundle size
- [x] Upgrade to React 19


tiagofilipenunes (approved):
Very good work, app looking snappier and snappier, approved 👍 

Can you please check the points below (non-blockers):
- [ ] Could you add a script in package.json to run the vite bundle analyzer?
- [ ] I would prefer to remove the compiler packages for now if possible, and instead create an issue explaining how to enable it and handle the compiler in a separate PR instead of having commented out code
- [ ] Since you also updated framer-motion, please check that this issue is now fixed #139.

Nice lazy loading highcharts, removing a lot of lodash uses. The waitFor in the e2e tests might improve their flakiness, nice catch too. Nice to see we can finally use Intl.Segmenter instead of Graphemer.

PR #1623, opened 2025-01-06
Portfolio Overview Dashboard

### LINKED ISSUE #1593: Portfolio Overview Dashboard
### Overview
We aim to provide users with a quick overview of their portfolio by displaying key metrics such as net worth, total trades, and trades made in the last 24 hours. This allows users to assess the status of their portfolio at a glance without needing to dive into the details of each strategy.

### Design
[Web](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=18792-4357&t=5ZtBaiApfdJhRrV9-1)
[Mobile](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=18792-3988&t=5ZtBaiApfdJhRrV9-1)

### Implementation
The key metrics to be displayed will focus on each user's portfolio and become visible once the user connects their wallet.

**Metrics to Display:**
1. **Net Worth** - The total budget of all strategies (active and inactive) within the user's portfolio. The net worth value will update automatically when the app's selected currency changes.
2. **Total Trades** - The cumulative number of trades across all strategies in the user's portfolio. This excludes trades from strategies that have been deleted or transferred to another account.
3. **Trades (Last 24 Hours)** - Similar to total trades but limited to trades executed within the past 24 hours.

GrandSchtroumpf:
fix #1593

Desktop
image
Mobile
image


ashachaf (approved):

src/components/explorer/ExplorerHeader.tsx
```diff
@@ -0,0 +1,333 @@
...
+    .splice(0, 3);
+
+  // If there are less than 3, pick the remaining best
+  if (list.length < 3) {
+    const remaining = Object.values(pairs)
+      .filter((pair) => !!pair.pairTrades_24h)
+      .sort((a, b) => b.pairTrades - a.pairTrades)
+      .splice(0, 3 - list.length);
+    list.push(...remaining);
+  }
```
tiagofilipenunes, line 33:
```suggestion
  // If there are less than 3, pick the remaining best
  if (list.length < 3) {
    const remaining = structuredClone(list).splice(0, 3 - list.length);
    list.push(...remaining);
  }
  }
```

src/components/explorer/ExplorerHeader.tsx
```diff
@@ -0,0 +1,333 @@
...
+    .splice(0, 3);
+
+  // If there are less than 3, pick the remaining best
+  if (list.length < 3) {
+    const remaining = trades
+      .filter((t) => !!t.strategyTrades_24h)
+      .sort((a, b) => b.strategyTrades - a.strategyTrades)
+      .splice(0, 3 - list.length);
+    list.push(...remaining);
+  }
```
tiagofilipenunes, line 66:
```suggestion
  // If there are less than 3, pick the remaining best
  if (list.length < 3) {
    const remaining = structuredClone(list).splice(0, 3 - list.length);
    list.push(...remaining);
  }
```

src/components/explorer/ExplorerHeader.tsx
```diff
@@ -0,0 +1,333 @@
...
+          </Link>
+          <Link
+            to="/trade/market"
+            className={buttonStyles({ variant: 'white' })}
+          >
+            Trade
+          </Link>
+        </div>
+      </article>
+      <article className="border-background-800 grid hidden flex-1 gap-8 rounded border-2 p-20 md:block">
```
tiagofilipenunes, line 115:
```suggestion
      <article className="border-background-800 hidden flex-1 gap-8 rounded border-2 p-20 md:grid">
```

src/components/explorer/ExplorerHeader.tsx
```diff
@@ -0,0 +1,333 @@
...
+                  >
+                    {formatter.format(trades)}
+                  </Link>
+                </td>
+              </tr>
+            ))}
+          </tbody>
+        </table>
+      </article>
+      <article className="border-background-800 grid hidden flex-1 gap-8 rounded border-2 p-20 lg:block">
```
tiagofilipenunes, line 172:
```suggestion
      <article className="border-background-800 hidden flex-1 gap-8 rounded border-2 p-20 lg:grid">
```

tiagofilipenunes (changes requested):
Great work, the explore and my strategies headers look great, the slide up animation looks amazing!

Notes:
- The MyStrategiesHeader and ExplorerHeader do not have a border as the mockup, flagging just in case this isn't what is expected
- When picking a different currency in the MyStrategiesHeader, the value didn't match the expected conversion, for example:
image
image

src/components/explorer/ExplorerHeader.tsx
GrandSchtroumpf: We need to sort back with only the item that has pairTrades_24h. This is why I need to filter & sort again.

tiagofilipenunes (approved):

src/components/explorer/ExplorerHeader.tsx
```diff
@@ -0,0 +1,333 @@
...
+import {
+  PairTrade,
+  Trending,
+  useTrending,
+} from 'libs/queries/extApi/tradeCount';
+import { Link } from 'libs/routing';
+import { CSSProperties, useEffect, useRef } from 'react';
+import { toPairSlug } from 'utils/pairSearch';
+
+const useTrendingPairs = (trending?: Trending) => {
```
zavelevsky, line 14: I would say that this is a helper function disguised as a hook - and refactor it into a simple function that accepts tokensMap and trending and returns the processed data. It doesn't really deal with UI state and can be easily set aside to be tested separately and easily.

src/components/explorer/ExplorerHeader.tsx
```diff
@@ -0,0 +1,333 @@
...
+  return { isLoading: false, data };
+};
+
+interface StrategyWithTradeCount extends Strategy {
+  trades: number;
+}
+const useTrendStrategies = (
+  trending?: Trending
+): { isLoading: boolean; data: StrategyWithTradeCount[] } => {
+  const trades = trending?.tradeCount ?? [];
```
zavelevsky, line 53: here too, the logic to extract IDs of trending strategies can be a standalone pure function - which is then passed to a query

src/components/explorer/ExplorerHeader.tsx
GrandSchtroumpf: I transformed it into a pure function. I didn't move it though because it's specific to this component.

PR #1635, opened 2025-01-23
Remove currency picker from SEI, CELO & BLAST

### LINKED ISSUE #1634: Remove Currency Picker
### Overview
Since the market rates provided by the backend for chains other than Ethereum lack prices in currencies other than USD, we will remove the option to select a currency on those chains.

- For the Ethereum chain, the date picker will remain available.
- On all other chains, the date picker option will be hidden from the menu.
-- Market rates on these chains will continue to be displayed in USD only.

GrandSchtroumpf:
fix #1634 

- [x] Remove currency picker except on ethereum
- [x] Migrate localstorage for currentCurrency when not on ethereum

Test case
- Open preview with sei
- Open menu on the top right, Currency picker is not available
- Open preview with ethereum
- Open menu on the top right, Currency picker is available


Costa769 (approved):

tiagofilipenunes (approved):
LGTM, migration works flawlessly, can you please add a description of the currencyMenu to the readme?

PR #1636, opened 2025-01-24
Keep start & end on reload & between navigation

### LINKED ISSUE #1611: [Simulator] Sync start/end with the chart
When we zoom on the chart it updates the URL with start & end. But when we reload we do not update the chart to reflect the state from the URL.
- pass the current state to the Graph
- make a writable state instead of a readable state for start & end

GrandSchtroumpf:
fix #1611

- Keep start & end on reload on `/trade`, `/strategy/edit`, `/simulate`
- Keep start & end between concentrated & recurring on `/simulate` & `/strategy/edit`

How to test: 
- Go to simulate or trade
- Zoom in the price chart
- Reload
- Price should remain the same


src/components/strategies/common/d3Chart/D3PriceHistory.tsx
```diff
@@ -101,33 +97,44 @@ const useZoom = (
...
-  return { transform, zoomRange };
+      const selection = select<SVGSVGElement, unknown>('#interactive-chart');
+      const scale = data.length / days;
+      const translateX = baseXScale(from)!;
+      const transition = selection.transition().duration(duration);
+      const transform = zoomIdentity.scale(scale).translate(-1 * translateX, 0);
+      zoomHandler?.transform(transition, transform);
+      return new Promise((res) => setTimeout(res, duration));
+    },
+    [behavior, data, dms.boundedWidth, zoomHandler]
```
tiagofilipenunes, line 134:
```suggestion
    [behavior, data.length, dms.boundedWidth, zoomHandler]
```

src/components/strategies/common/d3Chart/D3PriceHistory.tsx
```diff
@@ -176,17 +185,23 @@ export const D3PriceHistory: FC<Props> = (props) => {
     zoomBehavior = 'normal',
     onRangeUpdates,
   } = props;
+  // const init = useRef<boolean>(false);
```
tiagofilipenunes, line 188:
```suggestion
```

src/components/strategies/common/d3Chart/D3PriceHistory.tsx
```diff
@@ -211,34 +226,61 @@ export const D3PriceHistory: FC<Props> = (props) => {
...
-  const zoomFromTo = ({ start, end }: { start?: Date; end?: Date }) => {
+  const zoomFromTo = async ({ start, end }: { start?: Date; end?: Date }) => {
     if (!start || !end) return;
-    zoomRange(toUnixUTC(startOfDay(start)), differenceInDays(end, start) + 1);
+    setListenOnZoom(false);
+    const from = toUnixUTC(startOfDay(start));
+    const to = toUnixUTC(startOfDay(end));
+    await zoomRange(from, differenceInDays(end, start) + 1);
+    onRangeUpdates({ start: from, end: to });
+    setListenOnZoom(true);
```
tiagofilipenunes, line 243: If you want to see the listenOnZoom effect immediately, wouldn't a useRef be better here? The two set states will overwrite eachother to only do `setListenOnZoom(true)`

src/components/strategies/common/d3Chart/D3PriceHistory.tsx
```diff
@@ -211,34 +226,61 @@ export const D3PriceHistory: FC<Props> = (props) => {
...
   };
 
-  const zoomIn = (days: number) => {
-    zoomRange(data.at(days * -1)!.date.toString(), days);
+  const zoomIn = async (days: number) => {
+    setListenOnZoom(false);
+    await zoomRange(data.at(days * -1)!.date.toString(), days);
+    setListenOnZoom(true);
```
tiagofilipenunes, line 249: Same here

tiagofilipenunes (changes requested):
Greatly improves navigability between different types of strategies in the same page! 🎉 Just some minor comments. 

I know you're aware but just flagging that the e2e screenshots are scrolling down in my-strategy

Seeing an issue where sometimes the range isn't updated automatically when pressing the 7D/1M/3M/1Y interval selectors:



https://github.com/user-attachments/assets/8348b8b9-df10-4b5c-a772-03a441c0ab69

src/components/strategies/common/d3Chart/D3PriceHistory.tsx
GrandSchtroumpf: Not really because there is a `await` in the middle so React would have the time to rerender. The goal is to force the other useEffect to stop listening on zoom end event.

src/components/strategies/common/d3Chart/D3PriceHistory.tsx
GrandSchtroumpf: I cannot because I need to map over the dates:
```js
.domain(data.map((d) => d.date.toString()))
```

Technically I could (and I should I guess) because I know the array will always be the same... But so far I don't see any rendering issue right now, so I would rather not play with these yet.

ashachaf (approved):

zavelevsky (approved):
Code LGTM, didn't test.

PR #1668, opened 2025-02-07
Add batchSize & throttleTimeout for both wagmi & sdk

### LINKED ISSUE #1667: [Config] Set chunkSize & delay

GrandSchtroumpf:
fix #1667

- Add batchSize & throttleTimeout in config
- Use them for the SDK
- Use them also for wagmi provider
- Use Promise.allSettle for the missingTokens in case one network call fails


README.md
```diff
@@ -276,6 +276,8 @@ The file `common.ts` with type [`AppConfig`](src/config/types.ts) contains impor
 - `tokenLists`: List of token lists including the uri and the parser name to be used to parse the token list. Please update the tokenParserMap in the `src/config/utils.ts` file to include the parser name and the parser function.
 - `sdk`:
   - `cacheTTL`: When the app loads, it will ignore any cached data if it is older than the cacheTTL time in milliseconds. If set to 0, the app will always ignore the cache data and fetch new data on load.
+  - `batchSize`: Amount of rpc request per batch
```
zavelevsky, line 279: I find it confusing "Amount of rpc request per batch" - it's actually how many blocks in a range when reading logs

src/workers/sdk.ts
```diff
@@ -54,7 +58,13 @@ const init = async (
     chainId
   );
   api = new ContractsApi(provider, config);
-  const { cache, startDataSync } = initSyncedCache(api.reader, cachedData);
+  const { cache, startDataSync } = initSyncedCache(
+    api.reader,
+    sdkConfig?.cache,
+    sdkConfig?.batchSize,
```
zavelevsky, line 64: you are passing batch size in two place here - but the first place is actually how many pairs to include in a single multicall call to strategyByPair. Default in the SDK is 100.

README.md
```diff
@@ -276,6 +276,8 @@ The file `common.ts` with type [`AppConfig`](src/config/types.ts) contains impor
...
+  - `throttleTimeout`: timeout in millisecond between each request
```
zavelevsky, line 280: "timeout in millisecond between each request" ... to read latest events

src/workers/sdk.ts
```diff
@@ -54,7 +59,13 @@ const init = async (
     chainId
   );
   api = new ContractsApi(provider, config);
-  const { cache, startDataSync } = initSyncedCache(api.reader, cachedData);
+  const { cache, startDataSync } = initSyncedCache(
+    api.reader,
+    sdkConfig?.cache,
+    sdkConfig?.batchPairSize,
+    sdkConfig?.throttleTimeout,
```
zavelevsky, line 66: numOfPairsToBatch - number of pairs to include together in the same multicall when calling strategiesByPair. The bigger the number the less multicalls are required but on the other hand there's a risk to exceed RPC limitations when returning many strategies as a result.

msToWaitBetweenSyncs - related to reading latest events in order to keep up to date with chain state. This is the number of milliseconds to wait between syncs - so after a sync is completed, that is after we've done reading all the logs that took place in the blocks since the last processed one - we will wait this timeout before starting over.

chunkSize - also related to reading latest events in order to keep up to date with chain state. As the range of blocks since the last processed one and the latest one can be wide - breaking the range into chinks is required in most RPCs. This setting is the number of blocks to fetch in a single chunk.

src/workers/sdk.ts
```diff
@@ -42,7 +42,12 @@ const init = async (
   },
   config: ContractsConfig,
   decimalsMap?: Map<string, number>,
-  cachedData?: string
+  sdkConfig?: {
+    cache?: string;
+    batchPairSize?: number;
+    batchBlockSize?: number;
```
zavelevsky, line 48: I think batchPairSize and batchBlockSize are confusing names. If you disagree than at least it would be grammatically more correct to call it pairBatchSize and blockBatchSize.

README.md
```diff
@@ -258,7 +258,11 @@ The file `common.ts` with type [`AppConfig`](src/config/types.ts) contains impor
   - `defaultLimitedApproval`: Optional flag to set the default ERC-20 approval to limited approval. For chains where gas is low, it is recommended to set this flag to true.
   - `gasToken`: Gas token name, symbol, decimals, address and logoURI. This parameter will take priority over the `tokenListOverride`.
   - `blockExplorer`: The name and URL of the block explorer to be used in the notifications and when the network is added to the injected wallet.
-  - `rpc`: The RPC url and headers of the network, used to add the network to the injected wallet and to fetch data from the chain.
+  - `rpc`: 
+    - `url`: RPC url of the network
+    - `headers`: Headers append to each request to the RPC network
+    - `batchSize`: TThe maximum number of JSON-RPC requests to send in a batch
```
zavelevsky, line 264: "TThe maximum number of JSON-RPC requests to send in a batch" is not a correct description - please see my explanations.

README.md
```diff
@@ -258,7 +258,11 @@ The file `common.ts` with type [`AppConfig`](src/config/types.ts) contains impor
...
+    - `wait`: The maximum number of milliseconds to wait before sending a batch
```
zavelevsky, line 265: incorrect explanation - it's not between batches - it's between sync cycles.

README.md
GrandSchtroumpf: This is the description of the wagmi provider batch request. I copied their description

README.md
GrandSchtroumpf: This is the wagmi provider config. I copied their description

README.md
```diff
@@ -276,6 +280,9 @@ The file `common.ts` with type [`AppConfig`](src/config/types.ts) contains impor
 - `tokenLists`: List of token lists including the uri and the parser name to be used to parse the token list. Please update the tokenParserMap in the `src/config/utils.ts` file to include the parser name and the parser function.
 - `sdk`:
   - `cacheTTL`: When the app loads, it will ignore any cached data if it is older than the cacheTTL time in milliseconds. If set to 0, the app will always ignore the cache data and fetch new data on load.
+  - `pairBatchSize`: Amount of pairs batched together in the same multicall when calling strategiesByPair
+  - `blockRangeSize`: Amount of blocks since last call when syncing to latest logs
```
zavelevsky, line 284: Please change this:

```
  - `blockRangeSize`: Amount of blocks since last call when syncing to latest logs
```

into

```
- `blockRangeSize`: Max number of blocks to read logs from in a single call. Each time the SDK reads logs since last processed block it will split the total range into smaller ranges according to this configuration. This should be determined according to the RPC limitations.
```

zavelevsky (approved):

PR #1697, opened 2025-03-10
Full Range Strategy

### LINKED ISSUE #1693: Create/Edit strategy - add support for Full Range
## Goal
Carbon strategies can support a wide range of trading options which are not fully exposed in the app. one of these options is the ability to provide liquidity on the full price range of the token, similar to full range amm solutions with 1 unique advantage where the strategy creator can decide and set the spread with no restrictions.
The goal is to expose this capability on the app and allow users to access it more easily.

## Mockups
[Designs](https://www.figma.com/design/YpsatcY9jtxJr0SJwlFYNj/Carbon-Screen-Design?node-id=19615-47440&t=ZNYckJxgoUHuNaej-4)

## Specs
Under the existing "Concentrated" strategy creation flow, we would like to include a set of pre-defined options that can be used to set the prices automatically. One of these cases is "Full Range".

### Pre defined value
when to highlight the pre defined options?
- 0.5%, 1%, 10%, 20%: IF (currentMarketPrice*(100-preDefined)=minPrice) AND (currentMarketPrice*(100+preDefined) = maxPrice), highlight the relevant pre defined option
- Full Range: IF (currentMarketPrice/1000 = minPrice) AND (currentMarketPrice*1000 = maxPrice), highlight the relevant pre defined option

### Create Strategy
Under the "Price Range" section, where the user is requested to input the min and max prices to describe the range, please add the following pre-defined options.

IF user input/drag the min/max values manually
- this is the current behavior. no change is needed.
- IF user prices fit any pre-defined option, highlight that option as if it was selected (our default setup forces ±1% from price, in this case also highlight the pre-defined option that match this) 

IF selected pre-defined != "Full Range"
- when user selects any of the pre-defined options except "Full Range", apply this calculation on the prices
-- min price = current market price * (100% - pre-defined option) = $2550 * (100%-0.5%) = $2550 * 99.5% = $2,537.25
-- max price = current market price * (100% + pre-defined option) = $2550 * (100%+0.5%) = $2550 *100.5% = $2,562.75
- chart behaves as it is today (both historical and range)

IF selected pre-defined = "Full Range"
- Full Range pre-defined option
-- display minPrice = 0 in all relevant locations (input field, chart prices, and any other visible location). IF a user tries to interact with the 0 price input field, reset the display price in the field. if user exits with no changes, display 0 as before. if user changes the value, reset the ∞ input field as well (clear the input field and wait for user input).
-- display maxPrice = ∞ in all relevant locations (input field, chart prices, and any other visible location). IF a user tries to interact with the ∞ price input field, reset the display price in the field. if user exits with no changes, display ∞ as before. if user changes the value, reset the 0 input field as well (clear the input field and wait for user input).
-- contract minPrice value: currentTokenPrice / 1000
-- contract maxPrice value: currentTokenPrice * 1000
- chart visibility: [optional as we know how to handle it today], to improve visibility and make sure the spread is still visible and the historical price chart is not flatten, consider the following chart adjustments when a user selects the "Full Range" option
-- history chart: 
--- y-axis changes: keep the chart visibility at previous zoom (same y-axis min and max prices), however, replace the y-axis grid from showing prices, and replace it with ∞ at the top and 0 and the bottom of the y-axis. 
--- min/max price chart indication: remove the indicators that show the prices against the y-axis. this should also force a disable on the option to drag them up/down. 
--- remove chart buffer: currently, we apply a buffer between min/max price and bottom/top of the chart area. in Full Range, the colored area should reach the bottom and top of the chart area with no buffer.
-- range chart:
--- x-axis changes: similar to the historical chart, maintain the same grid zoom as before the Full Range selection in order to maintain the spread visibility, however, replace the x-axis prices to show 0 on the left and ∞ on the right. 
--- IF spread changes, chart should stay the same as it represents a Full Range anyways.
--- min/max price chart indication: remove the indicators that show the prices against the x-axis. this should also force a disable on the option to drag them left/right. 
--- remove chart buffer: currently, we apply a buffer (padding) between min/max price and right/left ends of the chart area. in Full Range, the colored area should reach the right and left ends of the chart area with no buffer.
- Warning/Error messages
-- user cannot input 0 or ∞ as input values in the price fields. we already have warnings for invalid prices/input so this is covered
- price input field supporting information
-- % from price indication: hide this indication
-- $ price indication: hide this indication
- Budget calculation: since the range is structured to be equal in both above and below the current market price, the existing calculation will work as expected for budgets. no change is needed.

### Edit Strategy
apply the same changes as the create strategy page.
- deposit/withdraw: apply the same logic as we apply today on concentrated liquidity
- editing into a Full Range: apply the same rules as create
- editing out of a Full Range: apply the required defaults according to the relevant strategy

### Simulate
apply the same changes as the create strategy page.
when running the simulation, it will use the "contract" min/max prices defined above.

### Strategy Overview Card
The overview card is designed to display the strategy details. In the Full Range case, we would like to apply some changes on how this information is shown

IF (sellHigh.maxPrice / buyLow.minPrice) >= 1,000,000 //since we apply factor of 1000 the max/min price should always result in 1million to indicate this is a full range strategy (as defined in the interface)
- chart display
-- replace the prices shown on the x-axis grid. display 0 as the left most value, and ∞ as the right most value.
-- [optional] maintain visibility of ± 1% above/below the spread. applied as buyLow.marginalPrice*(100%-1%) and sellHigh.marginalPrice*(100+1%)
- order tool tip - apply the Full Range visible values
-- minPrice = 0
-- maxPrice = ∞

#### Settings menu
- Duplicate: when clicked, skip the "undercut" popup and open the create strategy page with prices pre-defined option of "Full Range"
- Simulate: when clicked, open simulate with prices pre-defined option of "Full Range"
- Pause: set min and max prices to 0. as we do today. no change needed.
- Delete, Deposit, Withdraw, Edit Prices: all behave as today. no change needed.

GrandSchtroumpf:
fix #1693


ashachaf:
- [ ] strategy overview card
can you set the "market price" to be in the middle of the area?
Screenshot 2025-03-11 at 16 09 46

- [ ] strategy overview card tool tip
min price = 0
max price = ∞
image

zavelevsky (approved):

ashachaf:
- [ ] full range historical chart
can we lock the historical chart as well when selecting "full range"?
currently:
- select full range
- change view to historical chart
- drag the chart left/right
expected: the setting to stay as "full range"
actual: the setting switch to very low/high prices

- [ ] (optional) market price update
- select full range
- open "edit market price" 
- change the price
expected: for the "full range" low/high prices to be re-calculated based on the new price

ashachaf (approved):

PR #1802, opened 2025-06-06
implement ethers v6

### LINKED ISSUE #1799: Move to Viem or Ethers 6 to support EIP 7702
Ethers 5 doesn't support EIP 7702, we need to use either ethers 6 or viem.
Since we're using wagmi in the frontend (which is based on viem), we get rid of ethers 5 and use viem instead.

GrandSchtroumpf:
fix #1799


GrandSchtroumpf:
"we were seeing many calls being issued all the time - and it seems like before any request it was checking the network type - so we changed to static so it won't keep doing it."
Let's check if we're seeing this again.

zavelevsky (approved):
Code LGTM - things to test:
integration with wallets
calls to chain not preceded with calls to check chain id

PR #2063, opened 2026-01-30
Dex aggregator

### LINKED ISSUE #2058: Centralize approval logic
Currently the approval logic runs on every change. This is expensive and force complex logic to support both classic transaction with EIP7702.
Best would be to move the approval down the line, and trigger it once we know EIP7702 isn't an option.
A potential side effect would be to delay the modal opening (~500ms to 1s) as we need to check for approval at the end instead of all the time.

Once this is done we could have a centralize creation function (`useCreateStrategy`) in all places, including the cart and migration page.

### LINKED ISSUE #2070: Update Privacy text
please update the privacy page with the updated version found here: https://docs.google.com/document/d/1afjLLpNKWsekkMgZ7CbxdsxdCwN9-lXe/edit

the "last updated" date should be the date we push the changes to the app.

GrandSchtroumpf:
Integrate dex aggregator
fix #2058
fix #2070


zavelevsky (commented):
The code LGTM

ashachaf (approved):

PR #2118, opened 2026-03-19
Gradient

GrandSchtroumpf:
Cooked for more than a year in the most secret labs of Bancor. It's ready to take over the world of Defi by storm.
After the PR, nothing will ever be the same again.

Go to debug and set: 
- RPC URL: https://virtual.mainnet.eu.rpc.tenderly.co/137dc8f2-123e-42ec-95c8-14dd64358381
- API URL: https://carbon-multi-endpoint-staging-152368584642.europe-west2.run.app/v1/proxy/b4345496-2dee-41bb-bd5d-cd00d9e2633f/

Go to Explorer and filter by "Gradient" to see existing gradient strategies


ashachaf:
- [x] auction price presets
when clicking the price presets (below the price input)
expected: the relevant price (start/end) will be adjusted according to the preset and the line on the chart would also change accordingly
actual: only the input field change. the line seems to use "market price"

- [x] auction chart view
currently: 
image

updated: zoom in slightly and make the auction duration to take more of the right side.  start price might need to be set at 5% and not 10%.
image

- [x] input price preset calculation
auction buy -> price change is positive instead of negative
other auctions -> incorrect price change

seems that when selecting -5% preset, the price change is in the other direction (up and not down) and at much smaller amount +0.05% (should be 5%)
image

- [x] delta in quote token
the delta calculation is using the number as an addition to the sell prices. it should be deduction.
for example, if i put in 20, the start and end prices should be 20 usdc less than the sell prices.
image

- [x] default settings
i would like to maintain 2 default settings (if possible).
1. stable to stable 
- if the user selects any 2 of the following tokens: usdc, usdt
- default start price: buy=-0.2%, sell=+0.2%
- default end price: buy=-0.02%, sell=+0.02%
2. all the rest -> use the current default settings

ashachaf:
- [x] quick auction duration picker
menu is on the very right part of the screen (to the point it actually goes out of the screen to the right)
after selecting a time, i expect the drop menu to close
when i mouse over outside the drop down, i expect the drop menu to close
image

- [ ] create strategy error
trying to create "quick auction", https://gradient.carbon-app-csq.pages.dev/trade/quick-auction?base=0xEeeeeEeeeEeEeeEeEeEeeEEEeeeeEeeeeeeeEEeE&quote=0xA0b86991c6218b36c1d19D4a2e9Eb0cE3606eB48&startPrice=%222277.2010126703835%22&endPrice=%222033.1326582778%22&deltaTime=%2231%22&budget=%220.001%22&direction=sell&marginalPrice=%222277.2010126703835%22
(unchecked signer is marked in debug)
image

- [x] strategy card updates
for time base strategies, we would like to include indication for "future" orders that might appear later on.
for this, please only apply the following:
IF startTime>currentTime (meaning it is planned to start in the future), 
  IF buy low order is set correctly (with prices !=0 and budget >0) -> display the green hourglass. 
  IF sell high order is set correctly (with prices !=0 and budget >0) -> display the red hourglass.
IF startTime<currentTime<endTime (the order is active) -> display the line (as we do today). do no show any hourglass.
IF endTime<currentTime (the order has expired) -> do not display the line or hourglass. this represents an inactive order as if price =0.

hourglass mouse over should display the "future order" indication
image

gray hourglass should only be visible at the top of the card next to the state and id.

- [x] filter name
change "gradient" with "time based"
image

- [x] time based "marginal" price indication
expected: the "current marginal price is" to update based on the auction time. in this example below, i've set a 5min auction with 3000 usdc price change so every few seconds the marginal price should be different.
image

- [ ] homepage after wallet connect
we display up to 4 cards side by side on the same row (on my screen at least).
can we force max 3 cards per row?

- [x] quick channel
for some reason, i cannot pick a duration on the drop down menu.
image

- [x] channel - percentage option
this might be problematic as the distance between the two lines changes based on the price if we apply percentage. 
we might need to remove this option.

- [x] quick triangle
incorrect error/warning message.
when the orders cross each other, we want to warn with this error text
Sell and Buy auction prices cannot cross. This configuration may result in loss of funds.
current error message:
image

updated: 
image

GrandSchtroumpf:
@ashachaf 
- "when i mouse over outside the drop down, i expect the drop menu to close": Currently this behavior is nowhere in the app.
- "trying to create "quick auction": I cannot reproduce
- Should I display the line for future order ?
- "expected: the "current marginal price is" to update based on the auction time" it's now updating each time you over the line. Is it ok ?
- "can we force max 3 cards per row?": the amount of card depends on the space available. We can either make the card larger, but it'll make the chart much higher (aspect ratio is same), or we keep the same size have a white space on both side
Option 1 
image
Option 2
image
- "we might need to remove this option." I can calculate the delta value based on startPrice and apply it to start & end. Or remove the option. As you prefer

ashachaf:
- [x] strategy management menu
remove the div above the only option
image

- [x] refine the definition of active/inactive strategy
both orders are set to begin in the future.
this means this strategy is inactive at the moment.
image

ashachaf:
- [ ] explore (and portfolio) - recently/oldest created sorting
currently, we use the strategy id to identify old/new strategies and basically sort ascending/descending. 
when mixing 2 controllers, we now have "new" ids which are being pushed down the list and refer to as "old" due to the strategy id sorting.
update, use create time for this type of sort

ashachaf:
- [x] active strategy indicated as inactive
this strategy has an active buy order.
however, the visible state is "inactive".
https://gradient.carbon-app-csq.pages.dev/strategy/57896044618658097711785492504343953926975274699741220483192166611388333031432
image

- [x] active and inactive grouping
for static strategies, we display active strategies first and then all inactive ones.
for time bases, seems that we do not apply the same logic.

- [x] time based counter
not sure how viable it is, can we force all blinking to be synced together?
meaning, they all gray out at the same time when blinking.
https://github.com/user-attachments/assets/dea07270-48f5-4ca4-887a-839b6221c3f8

